PR #10, opened 2016-12-04
README was referencing a missing file generator.go.

guptasu:
Fixed it by compiling all the generator file together without the file name
Also made the output name explicit (for some reason the built file was getting named as 'classmodels' without the -o flag)


timburks:
Thanks for catching this - when I wrote the README, all of the generator code was in one file (the missing generator.go).

PR #25, opened 2017-02-06
Add Swift code generator plugin

timburks:
This set of commits adds a code generation plugin (written in Swift) that generates Swift code for API clients and servers for specified OpenAPI models. The plugin is not complete but intended as an illustration of a code generation plugin that could be written with Swift. This initial commit generates a generic client and a server that uses the Kitura web framework.

The plugins/go/openapi_swift_generator/examples/bookstore directory contains an example and instructions for demonstrating the plugin.


.travis.yml
```diff
@@ -1,12 +1,43 @@
...
+  - ./.travis-install.sh
 
 script:
   - go test . -v
-  - cd plugins/go/openapi_go_generator/examples/bookstore
+  - pushd plugins/go/openapi_go_generator/examples/bookstore
+  - make test
+  - popd
+  - export PATH=.:$HOME/local/bin:$PATH
+  - export LD_LIBRARY_PATH=$HOME/local/lib
```
guptasu: Should this be done inside the -install.sh script ?

.travis-install.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
```
guptasu: Should we put all this inside the before_install step inside the original .travis.yml file ?

plugins/swift/openapi_swift_generator/Sources/openapi_swift_generator/InternalLoader.swift
```diff
@@ -0,0 +1,51 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Stencil
+import Foundation
+
+// A class for loading Stencil templates from compiled-in representations
+
+public class InternalLoader: Loader {
```
guptasu: TemplateLoader ?

plugins/swift/openapi_swift_generator/Sources/openapi_swift_generator/main.swift
```diff
@@ -0,0 +1,42 @@
...
+  let request = try Openapi_Plugin_V1_Request(protobuf: rawRequest)
+  let wrapper = request.wrapper
+  let document = try Openapi_V2_Document(protobuf:wrapper.value)
+
+  // build the service renderer
+  let renderer = ServiceRenderer(document:document)
+
+  // generate the desired files
+  var response = Openapi_Plugin_V1_Response()
+  let filenames = ["client.swift", "service.swift", "types.swift"]
```
guptasu: Any reason why this is different than the other sample where client and server were conditionally generated ?
https://github.com/googleapis/openapi-compiler/blob/master/plugins/go/openapi_go_generator/main.go#L51

.travis.yml
timburks: IIRC, I tried that and the values weren't set outside of the script.

.travis-install.sh
timburks: I will try that.

plugins/swift/openapi_swift_generator/Sources/openapi_swift_generator/InternalLoader.swift
timburks: That would be fine. Here's some [background on that](https://github.com/kylef/Stencil/issues/97). Stencil has an alternate loader called FileSystemLoader that reads from the file system. I think "InternalLoader" would be a good name for use inside Stencil, but see that it's confusing in our code.

plugins/swift/openapi_swift_generator/Sources/openapi_swift_generator/main.swift
timburks: Laziness. I'll add that.

timburks: Done [here](https://github.com/googleapis/openapi-compiler/pull/25/commits/a31fd29f3f6060810c6a3fd5dadff4a5b5e66d4d).

plugins/swift/openapi_swift_generator/Sources/openapi_swift_generator/InternalLoader.swift
timburks: Done [here](https://github.com/googleapis/openapi-compiler/pull/25/commits/9f9913bdac82266a55f19ec9095a57ea6c34f427).

.travis-install.sh
timburks: Can we set environment variables in the steps under install or before_install?

PR #38, opened 2017-05-05
Prohibit additional properties in schema definitions

tfesenko:
Prohibit additional properties in schema definitions, otherwise, the JSON schema is too permissive. 
OpenAPI v3 specification provides [specification extensions](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#specificationExtensions) to define additional (custom) properties.

## False validation errors with current JSON Schema
A too permissive schema can also cause false negatives (false validation errors) on a valid OpenAPI v3 specification. Let's take this example, it's simplified version of [PetStore](https://github.com/OAI/OpenAPI-Specification/pull/985/files#diff-4530232d2a5e2e4be11ab57cd8ba8894):
```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
paths: {}

components:
  schemas:
    Pet:
      properties:
        name:
          type: string
    Pets:
      type: array
      items:
        $ref: "#/components/schemas/Pet"
```
Current JSON Schema will cause an error on line 18 (`$ref: "#/components/schemas/Pet"`):
```
	Failed to match exactly one schema:
	 - /definitions/schemaOrReference/oneOf/0:
	 - /definitions/schemaOrReference/oneOf/1:
```
The problem is that it matches **both** `#/definitions/schema` and `#/definitions/reference` schemas, while exactly **one** match is allowed in this case. And our `$ref` matches `#/definitions/schema` because the `#/definitions/schema` definition did not prohibit additional properties. Similarly in other cases.

--------------------------------------
Thank you for creating this JSON Schema! #

PR #40, opened 2017-05-15
Fix regex in patternProperties

tfesenko:
[Based on the openapi-v3.0.0-rc2 branch]

Generated JSON Schema uses the following values as regex in patternProperties:
* `{name}`
* `{path}`
* `{media-type}`
* `{expression}`
* `{property}`

They are human-friendly and easy to understand, but unfortunately, JSON Schema validator does not treat it the same way as humans. 

## `{` and `}` in patternProperties
[In JSON Schema](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.17):
> Each property name of this ("patternProperties") object SHOULD be a valid regular expression, according to the ECMA 262 regular expression dialect.  

ECMA 262 does not allow using `{` and `}` as a PatternCharacters:
> PatternCharacter ::
SourceCharacter but not one of -
^ $ \ . * + ? ( ) [ ] { } |

and treats their combination as a [Quantifier](https://www.ecma-international.org/ecma-262/5.1/#sec-15.10.2.7) (range for matching the patterns).

Some JSON Schema validators fail on such values for pattern properties, others simply don't match elements as expected.

## Additional information regarding `{name}` and `{path}` in OpenAPI v3 specification
So, maybe we should escape`{` and `}` in pattern properties? Not really.
OpenAPI v3 specification provides some insights about what the component `{name}`s [should look like](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#componentsObject):
> Component names should match "^[a-zA-Z0-9\.\-_]+$"

And as for [`{path}`](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#paths-object):
> The field name MUST begin with a slash

JSON Schema for OpenAPI v2 [uses "^/"](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/schemas/v2.0/schema.json#L173) as regex for paths - I used the same regex for v3.

I used an "allow anything" (`^`) regex for remaining pattern properties (`{media-type}`, `{expression}`, and `{property}`).

### DIFF
--- a/OpenAPIv3/openapi-3.0.json
+++ b/OpenAPIv3/openapi-3.0.json
@@ -153,7 +153,7 @@
       "type": "object",
       "description": "",
       "patternProperties": {
-        "{name}": {
+        "^[a-zA-Z0-9\\.\\-_]+$": {
           "$ref": "#/definitions/serverVariable"
         },
         "^x-": {
@@ -230,7 +230,7 @@
       "type": "object",
       "description": "Holds the relative paths to the individual endpoints and their operations. The path is appended to the URL from the `Server Object` in order to construct the full URL.  The Paths MAY be empty, due to ACL constraints.",
       "patternProperties": {
-        "/{path}": {
+        "^/": {
           "$ref": "#/definitions/pathItem"
         },
         "^x-": {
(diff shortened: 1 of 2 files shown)


tfesenko:
Dear @googlebot, I signed it!

MikeRalphson:
Merging #38 causes my validator to fail against the new schema, but I believe merging this on top will at least partially resolve the issue. In addition, I think the regex for response codes should be widened from `^([0-9]{3})$` to `^([0-9X]{3})$` to allow for [response ranges](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#patterned-fields-4). Great work @tfesenko

Issue #42, opened 2017-05-17
OpenAPI 3 schema - scopes patternProperties regex is too restrictive


MikeRalphson:
Paging @tfesenko and @timburks 

The regular expression for scope names has been set to `^[a-zA-Z0-9\\.\\-_]+$` which matches what `{name}` is set to for elements under the `components` property.

I believe this is too restrictive in this case. As far as I can see scope names are only limited by their definition in [RFC6749](https://tools.ietf.org/html/rfc6749#section-3.3).

I'm not sure how JSON schema patternProperties disambiguation is supposed to work when more than one regex matches (in this instance `^x-` would also match a valid scope name.

PR #43, opened 2017-05-18
Amend Scopes and HTTP Response Code regexes

MikeRalphson:
refs #42 

Please forgive me, these are the first lines of go I've ever written and I haven't done any reading on the language. Happy to rev the PR.

### DIFF
--- a/OpenAPIv3/openapi-3.0.json
+++ b/OpenAPIv3/openapi-3.0.json
@@ -549,7 +549,7 @@
       "description": "A container for the expected responses of an operation. The container maps a HTTP response code to the expected response.  It is not expected for the documentation to necessarily cover all possible HTTP response codes, since they may not be known in advance. However, it is expected for the documentation to cover a successful operation response and any known errors.  The `default` MAY be used as a default response object for all HTTP codes  that are not covered individually by the specification.  The `Responses Object` MUST contain at least one response code, and it  SHOULD be the response for a successful operation call.",
       "additionalProperties": false,
       "patternProperties": {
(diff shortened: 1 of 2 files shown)


MikeRalphson:
I signed it!

timburks:
Thanks @MikeRalphson. 

Do you think we should have stricter regexes for paths (^/) and extensions (^x-)?

MikeRalphson:
> Do you think we should have stricter regexes for paths (^/) and extensions (^x-)?

I found just one [regex](http://www.regexlib.com/REDetails.aspx?regexp_id=2888) for matching the actual contents of a URL path (rather than extracting one):

```regex
^(?:(?:\.\./)|/)?(?:\w(?:[\w`~!$=;\-\+\.\^\(\)\|\{\}\[\]]|(?:%\d\d))*\w?)?(?:/\w(?:[\w`~!$=;\-\+\.\^\(\)\|\{\}\[\]]|(?:%\d\d))*\w?)*(?:\?[^#]+)?(?:#[a-z0-9]\w*)?$
```
though this includes the query and fragment parts which would need to be removed. I can't vouch for its correctness, and I must confess I have [reasons](https://github.com/APIs-guru/openapi-directory/blob/master/APIs/amazonaws.com/s3/2006-03-01/swagger.yaml#L106) for not wanting this to be overly strict if the OAS adopts this schema as their official one. I wouldn't object to a PR against the OAS to add a regex requirement though.

The OAS seems to put no other restrictions on extension names, any Unicode string is a valid JSON/YAML property name if suitably escaped.

PR #45, opened 2017-05-19
Copy patternProperties from parameter to header object

MikeRalphson:
As the `header` object takes its structure from the `parameter` object, we must copy the `patternProperties` as well as the `properties`, i.e. we make it extendable with specification extensions.

Similar disclaimer to before; go novice.


timburks:
LGTM

PR #52, opened 2017-06-05
Header object key regex was too restrictive

MikeRalphson:
The standard `name` regex, as used for re-usable header objects in the `#/components/headers` object is too restrictive for use in the `headers object` keys themselves.

As per [RFC7230](http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.2.6) several additional characters are allowed in header names.


timburks:
Would you double-check this now? It appears that the headers object is gone now (replaced by a map), so I'm not sure how we should use this new regex... maybe it's not needed.

MikeRalphson:
Hi @timburks - looking at the latest schema I can see:

* there is a headersOrReferences map 
* there is a headers map used by the requestBody/encoding property (should this also be headersOrReferences -possible defect in the spec?)
* we are not yet removing the `name` and `in` properties of the copied parameter object definition
* there is no restriction on the naming of the keys of headersOrReferences (additionalProperties is used instead of patternProperties)
* the rule for objects under `components` is that they must satisfy the regex `^[a-zA-Z0-9\.\-_]+$`
* however that rule does not apply **outside** components
* therefore it is currently problematic to use one map definition for all headers maps

I can raise an issue on the OAS repo asking for clarity...

timburks:
@MikeRalphson, thanks for reviewing this! 

@tfesenko asked if the headers map in [the encoding object](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#encodingObject) could contain references as a side question in [this issue](https://github.com/OAI/OpenAPI-Specification/pull/1126), but I suppose this change hasn't been made yet.

I've updated the schema builder to [remove the `name` and `in` properties from the header object](https://github.com/googleapis/gnostic/commit/9dee95c68c2e903ab86283cc829e88e9607c7142).

I hadn't noticed the regex restriction on objects under components but can add that, which will also change "additionalProperties" to "patternProperties" for those objects. 

How to represent the other maps is a question worth discussing. Should we replace additionalProperties with patternProperties for all maps and use ".*" for maps with unrestricted keys? Are there any other maps that will have key restrictions?

MikeRalphson:
Thanks @timburks I knew I'd seen a mention of the encoding header issue somewhere.

`name` and `in` removal LGTM.

I wonder if `additionalProperties` might perform better for schema validators than `patternProperties` with a universal regex? I think at this late stage it is unlikely that any other key restrictions will be added. 

There is little risk with people defining headers that break the RFC7230 spec, as the API implementation would probably not be able to send them.

PR #57, opened 2017-07-28
Reduce the amount of memory and CPU used by generated by OpenAPI code

smarterclayton:
The majority of uses of regex inside of the generated code suffer from easy to fix problems:

1. Use of uncompiled patterns - there are no patterns that aren't known ahead of time
2. Use of regex for trivial string operations.

This PR fixes kubernetes/kubernetes#49776 by removing the impact of the two issues above at
generation time.

In production systems using the generated code (Kubernetes and OpenShift) we measured that 1/3rd of all startup allocations and the majority of startup time was being consumed by this code (which is being used to parse and display OpenAPI models).  After this fix, allocations from this package drop to effectively zero.


timburks:
Awesome, thanks!

PR #65, opened 2017-11-01
Init a surface if a plugin runs as standalone program

orsobruno:
If we try to run a plugin as standalone program it will print the error and stops the execution. It happens because the model is not initialised.  

I added the initialization, moved the second condition inside else block to hold last error and added intelliJ IDEA project files to gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+# IntelliJ IDEA
+.idea
 # Eclipse
 .checkstyle
 .project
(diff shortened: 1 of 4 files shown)


orsobruno:
There is a bug in gnostic-go-generator plugin. If request method is POST and body has not parameters it throws the error: 

> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1268347]

How to reproduce:

`gnostic --go-client-out=outdir examples/v2.0/json/petstore.json`

PR #79, opened 2018-02-15
Add linter plugin to check for array responses

mkistler:
This PR adds a Node-based linter to check operation responses and flag any that are top-level arrays.

### DIFF
--- a/linters/node/gnostic-lint-responses/package.json
+++ b/linters/node/gnostic-lint-responses/package.json
@@ -1,0 +1,18 @@
+{
+  "name": "gnostic-lint-responses",
+  "version": "1.0.0",
+  "description": "Gnostic linter plugin to check responses",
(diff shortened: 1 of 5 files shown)


linters/node/gnostic-lint-responses/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "lodash": "^4.17.5",
+    "nexe": "^2.0.0-rc.24",
+    "protobufjs": "^6.8.4"
+  },
+  "devDependencies": {},
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC"
```
timburks: For consistency with the overall project, can we replace "ISC" with "Apache-2.0" here?

PR #84, opened 2018-03-15
OpenAPI v3 schema items value MUST be an object and not an array.

### LINKED ISSUE #83: OpenAPI 3: schema items - Value MUST be an object and not an array
Quote from [OpenAPI 3 Schema Object specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schema-object)

> items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema. items MUST be present if the type is array.

But now, in gnostic v3 model Items are array, what is make usage less obvious. 
Seems gnostic use only first item only anyway.

skipor:
Fixes #83


timburks:
Thank you!

PR #89, opened 2018-04-19
Move generator plugin samples to the "generators" top-level directory.

timburks:
This fixes problems with recursive gets of the "plugins" directory, i.e.
```
go get github.com/googleapis/gnostic/plugins/...
```


timburks:
The generators should go in separate repos.

PR #95, opened 2018-10-19
fix silent plugin failures & goimports dep

noahdietz:
* `goimports` wasn't explicitly installed by root `make build`, but is exec'd by `gnostic-go-generator` plugin (it is installed by the plugin's target).
* plugins were failing silently because `plugins.HandleResponse` return value (the potential plugin error) was being ignored

### DIFF
--- a/.travis-install.sh
+++ b/.travis-install.sh
@@ -14,7 +14,7 @@ cd
 mkdir -p local
 
 # Install swift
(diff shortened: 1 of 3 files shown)


noahdietz:
We might look to refactor the exec of goimports to the use of the Go API, I think this is it: https://godoc.org/golang.org/x/tools/imports

noahdietz:
I don't understand the errors in Travis, could use some help. `go test` passes locally (and it looks like in travis as well)

noahdietz:
Travis failures are due to broken dependency in generated Swift code. After talking offline, we will merge this PR and fix the broken dependency in generated Swift code in a separate one.

timburks:
+1 to the goimports refactor and removing Swift tests from gnostic travis.

This time the tests were broken by Swift version changes. If we restore these tests, it should be in a separate repo for the swift plugin.

PR #97, opened 2018-10-19
refactor goimports exec to imports pkg

noahdietz:
* refactor `renderer.go` to process imports & formatting with `imports` pkg
* remove `goimports` binary installation
* remove `goimports.go`

From discussion in #95


extensions/extension.pb.go
```diff
@@ -1,24 +1,12 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // source: extension.proto
 
-/*
```
noahdietz: I think this was included by accident. I don't think this should be included. Should I remove it?

timburks: Let's keep it. It looks like this changed because the protoc-gen-go output changed. I tested and an extension still built with the change, and extension support is still experimental (and incomplete).

Issue #99, opened 2018-11-27
no methods to clear or disable the fileCache and infoCache


samshan08:
when using gnostic to parse the openAPIv3 document, our code is as follow:
```
func unmarshalApiSpec(svcName string, apiSpecStr string) (*openapi_v3.Document, error) {
	decodeValue, err := Base64Decode(apiSpecStr)
	if nil != err {
		return nil, err
	}
	rawText, err := DeCompressDefinition(decodeValue)
	if nil != err {
		return nil, err
	}
	info, err := compiler.ReadInfoFromBytes(svcName, rawText)
	if err != nil {
		return nil, err
	}
	doc, err := openapi_v3.NewDocument(info, compiler.NewContext("$root", nil))
	if err != nil {
		return nil, err
	}
	return doc, nil
}
```
 We encounter the problem that: when we call unmarshalApiSpec twice using the same svcName, but different apiSpecStr, it is always return the first parse result unless you restart the process.
As we study the code, we found that gnostic has cache for cache the unmarshal result, as follow:
```
func ReadInfoFromBytes(filename string, bytes []byte) (interface{}, error) {
	initializeInfoCache()
	cachedInfo, ok := **infoCache[filename]
	if ok {
		if verboseReader {
			log.Printf("Cache hit info for file %s", filename)
		}
		return cachedInfo, nil
	}
	if verboseReader {
		log.Printf("Reading info for file %s", filename)
	}
	var info yaml.MapSlice
	err := yaml.Unmarshal(bytes, &info)
	if err != nil {
		return nil, err
	}
	if len(filename) > 0 {
		infoCache[filename] = info
	}
	return info, nil
}
```
We did not find any methods to clear or disable this cache.

timburks:
Fixed with #103

PR #101, opened 2018-11-30
improve: add methods to delete and disable caches

andrewshan:
add methods to delete and disable the cache, related to the issue #99


andrewshan:
I signed it!

PR #103, opened 2018-12-14
improvement for cache apis

andrewshan:
fixed for #99
- when cache is disabled, the result should not be added into cache
- add full unit test for the 4 apis


timburks:
Thanks for adding this!

Issue #111, opened 2019-02-08
gnostic-go-generator: -output doesn't like non package names


bhenderson:
I know the docs say that -out is also the package name, but the error message is confusing. when I do something like `-output=/tmp/bookstore` I get

```
Plugin error: [constants.go:3:9: expected 'IDENT', found '/']
```
I think it would be nice to have a friendlier error.

I've traced this to goimports errors. related to #110 
https://github.com/googleapis/gnostic/blob/d55a06a32dc7468c645ac01d93d37d4dd47ca062/plugins/gnostic-go-generator/renderer.go#L63

I made some changes like

```
diff --git a/plugins/gnostic-go-generator/renderer.go b/plugins/gnostic-go-generator/renderer.go
index 93fd47d..31efb8e 100644
--- a/plugins/gnostic-go-generator/renderer.go
+++ b/plugins/gnostic-go-generator/renderer.go
@@ -60,7 +60,9 @@ func (renderer *Renderer) Render(response *plugins.Response, files []string) (er
                }
                // run generated Go files through imports pkg
                if filepath.Ext(file.Name) == ".go" {
-                       file.Data, err = imports.Process(file.Name, file.Data, nil)
+                       if file.Data, err = imports.Process(file.Name, file.Data, nil); err != nil {
+                               response.Errors = append(response.Errors, fmt.Sprintf("ERROR %v", err))
+                       }
                }
                response.Files = append(response.Files, file)
        }
```

The last error gets reported twice though due to named returns, and I was thinking it would be really nice to still write out the files so that the user can inspect the problem in context. 

thoughts?

bhenderson:
The code I put above is really for better error messaging.

I'm working on a PR for cleaning the package name.

bhenderson:
Closed by #112

PR #112, opened 2019-02-09
Resolve package name from output path

bhenderson:
I wanted to be able to run -output . or -output /path/to/dir

Using format.Source was the easiest way I could figure out how to
validate the package name.

Relates to #111

### DIFF
--- a/plugins/gnostic-go-generator/main.go
+++ b/plugins/gnostic-go-generator/main.go
@@ -78,6 +78,8 @@ func main() {
 	env.RespondAndExitIfError(err)
 }
 
+// resolvePackageName converts a path to a valid package name or
+// error if path can't be resolved or resolves to an invalid package name
 func resolvePackageName(p string) (string, error) {
 	p, err := filepath.Abs(p)
 	if err == nil {
--- a/plugins/gnostic-go-generator/main.go
+++ b/plugins/gnostic-go-generator/main.go
@@ -19,6 +19,8 @@ package main
 import (
 	"encoding/json"
 	"errors"
+	"go/format"
+	"path/filepath"
 	"strings"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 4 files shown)


bhenderson:
travis build broke because it's now using go1.11.5 instead of go1.10.7 and apparently the behavior of go get changed. I'm looking into a fix. I think all it needs is the gnostic run to happen earlier.

plugins/gnostic-go-generator/main.go
```diff
@@ -74,3 +77,15 @@ func main() {
 	err = errors.New("No generated code surface model is available.")
 	env.RespondAndExitIfError(err)
 }
+
+func resolvePackageName(p string) (string, error) {
+	p, err := filepath.Abs(p)
```
noahdietz: Can you add some comments in here to describe the resolution behavior?

noahdietz:
Thanks @bhenderson for the PR. Apologies for the slow reply on this. 1 quick comment

plugins/gnostic-go-generator/main.go
bhenderson: how's that?

plugins/gnostic-go-generator/main.go
```diff
@@ -78,6 +78,8 @@ func main() {
 	env.RespondAndExitIfError(err)
 }
 
+// resolvePackageName converts a path to a valid package name or
+// error if path can't be resolved or resolves to an invalid package name
```
noahdietz: sorry for the nit but please add a `.` at the end :)

plugins/gnostic-go-generator/main.go
bhenderson: I'm not on my normal computer, so I'm just using the GH editor. If you want me to squash the commits I can do that tonight.

noahdietz: I'll squash them when I merge, thanks though!

Issue #110, opened 2019-02-07
lots of issues trying to generate an api


bhenderson:
Hi, I'm super excited to use this pkg, but I'm having lots of issues. I'd also like to help where I can.

I'm using this open api v3 spec https://us-east-1-production-zcv2-api.ziff.ai/_schema

first problem is:
```
ERROR $root.components.schemas.cloud.schema.FileEncoder.schema.properties.encoderModelUri.schema has invalid property: $ref
ERROR $root.components.schemas.cloud.schema.FileEncoder.schema.properties.encoderModelUri.reference has invalid property: readOnly
ERROR $root.components.schemas.cloud.schema.FileEncoder.reference is missing required property: $ref
ERROR $root.components.schemas.cloud.schema.FileEncoder.reference has invalid properties: type, required, properties
```

As I wasn't using that model, I just removed that one field in the json spec.

My next issue was

```
2019/02/07 09:22:57 unimplemented: any_of:<schema:<type:"string" > > any_of:<schema:<type:"integer" format:"int64" > > any_of:<schema:<type:"number" format:"double" > >
2019/02/07 09:22:57 unimplemented: any_of:<schema:<type:"integer" format:"int64" > > any_of:<schema:<type:"string" > >
```

I understand that anyOf types would be hard to implement in go, I just changed these to type string

`gnostic ziff-api.json --go-generator-out=ziff` then ran without errors but types.go and client.go were completely empty.

bhenderson:
I figured out there were goimports errors being silenced. I'm trying to look into the other things.

jbltx:
Any updates on this issue?

I am currently trying to convert an OpenAPI v3 specification and here's the result : 

```ps
$ gnostic.exe --resolve-refs --go-generator-out=toto spec.json
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x8 pc=0x839191]

goroutine 1 [running]:
main.(*Renderer).RenderClient(0xc000095eb0, 0x9, 0x91e658, 0x9, 0x0, 0x0)
        C:/Users/xxx/go/pkg/mod/github.com/googleapis/gnostic@v0.2.0/plugins/gnostic-go-generator/render_client.go:113 +0xfe1
main.(*Renderer).Render(0xc000095eb0, 0xc00009e5a0, 0xc000095f38, 0x5, 0x5, 0xc000098328, 0x0)
[...]

PR #113, opened 2019-02-11
Fix some edge cases

bhenderson:
Fixing a couple of problems with my api (#110). I can work on adding tests if you want to point me in the right direction.


bhenderson:
hrm, found a bug with the type identifier fix.

bhenderson:
just to be clear, I pushed a fix to the identifier bug.

PR #117, opened 2019-05-08
fix(resp): fix connection leak problem of missing response body close

liubog2008:
Body should always be closed even if status code is not 200


liubog2008:
@timburks

timburks:
Thank you!

liubog2008:
Could you release a new version?

liubog2008:
@timburks

PR #120, opened 2019-05-15
sort parameters for go client renderer

glickbot:
go client generator listed the parameters randomly such that if you generated the same client code twice, the parameters could be out of order. Added sorting to the ParameterList function to fix this.


plugins/gnostic-go-generator/render_client.go
```diff
@@ -15,20 +15,22 @@
...
+	"sort"
 	"strings"
 
 	surface "github.com/googleapis/gnostic/surface"
 )
 
 // ParameterList returns a string representation of a method's parameters
 func ParameterList(parametersType *surface.Type) string {
-	result := ""
+	parameters := make([]string, len(parametersType.Fields))
```
noahdietz: Please move the initialization of `parameters` inside the `nil` check of `parametersType`. There is an opportunity for a `nil pointer dereference`. It's failing CI as a result.

Thanks!

noahdietz: Ah but then the resulting sort and join wouldn't have a reference...

Let's check early instead with 
```go
if parametersType == nil {
  return ""
}
```

glickbot: There's a few changes I need to add too since it changes the bookstore tests. Might look into sorting the list if the user specifies an option to for backwards compatibility.

glickbot: Added an environment variable, mentioned it in the README, and moved checking for nil pointers earlier in the function.

plugins/gnostic-go-generator/render_client.go
```diff
@@ -22,14 +22,19 @@ import (
 )
 
 // ParameterList returns a string representation of a method's parameters
-func ParameterList(parametersType *surface.Type) string {
+func (renderer *Renderer) ParameterList(parametersType *surface.Type) string {
+	if parametersType == nil ||
+		parametersType.Fields == nil ||
+		len(parametersType.Fields) == 0 {
```
noahdietz: This isn't really a blocker, but would it be possible to make the following code work even when `len(Fields)` is `0`? Removing the need for this check. I like it when loop code works in 0 cases as well.

The thing I see preventing that is the extra `\n` appended to the resulting string...

plugins/gnostic-go-generator/main.go
```diff
@@ -36,6 +37,8 @@ func main() {
 	packageName, err := resolvePackageName(env.Request.OutputPath)
 	env.RespondAndExitIfError(err)
 
+	_, sortClientParamerList := os.LookupEnv("GNOSTIC_GO_SORTCLIENTPARAMETERLIST")
```
noahdietz: I'm not sure that an environment variable is the best way to go here. `gnostic` is modeled after `protoc`, which enables plugins to take parameters/options via a plugin option, for example, if this plugin were compatible with protoc: `--go-generator-opt="sort-params"`.

Can we add this feature in a similar way? Maybe `--go-generator-sort-params` and if the `Invocation` contains this flag, we set our `renderer` toggle to `true`?

plugins/gnostic-go-generator/render_client.go
```diff
@@ -15,20 +15,27 @@
...
+		return ""
+	}
+	parameters := make([]string, len(parametersType.Fields))
+	for i, field := range parametersType.Fields {
+		parameters[i] = field.ParameterName + " " + field.NativeType + ","
+	}
```
noahdietz: line break please

plugins/gnostic-go-generator/render_client.go
```diff
@@ -15,20 +15,27 @@
...
+	if renderer.SortParameterList {
+		sort.Strings(parameters)
 	}
-	return result
+	return strings.Join(parameters, "\n") + "\n"
```
noahdietz: line break before, please

plugins/gnostic-go-generator/render_client.go
```diff
@@ -15,20 +15,27 @@
...
-	if parametersType != nil {
-		for _, field := range parametersType.Fields {
-			result += field.ParameterName + " " + field.NativeType + "," + "\n"
-		}
+func (renderer *Renderer) ParameterList(parametersType *surface.Type) string {
+	if parametersType == nil ||
+		parametersType.Fields == nil ||
+		len(parametersType.Fields) == 0 {
+		return ""
+	}
```
noahdietz: line break please

plugins/gnostic-go-generator/renderer.go
```diff
@@ -26,14 +26,16 @@ import (
...
+	Model             *surface.Model
+	Package           string // package name
+	SortParameterList bool
 }
 
 // NewServiceRenderer creates a renderer.
 func NewServiceRenderer(model *surface.Model) (renderer *Renderer, err error) {
 	renderer = &Renderer{}
 	renderer.Model = model
+	renderer.SortParameterList = false // default
```
noahdietz: I'm going to ask you to change existing code to improve the style if you don't mind. Not sure why this is has an `error` in the return type either

```go
return &Renderer{ 
  Model: model,
  SortParams: false,
}, nil
```

plugins/gnostic-go-generator/renderer.go
```diff
@@ -26,14 +26,16 @@ import (
 
 // Renderer generates code for a surface.Model.
 type Renderer struct {
-	Model   *surface.Model
-	Package string // package name
+	Model             *surface.Model
+	Package           string // package name
+	SortParameterList bool
```
noahdietz: Can we shorten this name? `SortParams` or something similar.

.gitignore
```diff
@@ -19,3 +19,18 @@ bundle.json
 *~
 Package.resolved
 extensions/sample/generated
+
+plugins/gnostic-go-generator/examples/v3.0/bookstore/service/service
```
noahdietz: Was there an issue with these? I think we should keep these updated as long as they are checked in.

noahdietz:
I agree, that sounds like a better solution and use of time! Feel free to close this if you want to shift approaches

Issue #121, opened 2019-05-16
gnostic-go-generator isn't handling errors from import.Process in renderer 


glickbot:
gnostic-go-generator isn't handling errors from import.Process in renderer

https://github.com/googleapis/gnostic/blob/c273e2ca9014e3df36dc9e3a88b9c7c064610980/plugins/gnostic-go-generator/renderer.go#L63

glickbot:
Issue fixed via #122

PR #122, opened 2019-05-16
added error handling to golang plugin renderer

glickbot:
fix for #121


plugins/gnostic-go-generator/renderer.go
```diff
@@ -61,6 +61,9 @@ func (renderer *Renderer) Render(response *plugins.Response, files []string) (er
 		// run generated Go files through imports pkg
 		if filepath.Ext(file.Name) == ".go" {
 			file.Data, err = imports.Process(file.Name, file.Data, nil)
+			if err != nil {
+				response.Errors = append(response.Errors, fmt.Sprintf("ERROR %v", err))
```
noahdietz: Let's just append the raw error string, no need to prefix it with "ERROR"

Issue #125, opened 2019-05-23
surface model and go plugin doesn't handle 'any' type


glickbot:
surface model and go plugin doesn't handle 'any' type
example:

disco get bigquery v2 --openapi3
gnostic openapi3-bigquery-v2.pb --go-client-out=bigquery
2019/05/23 15:38:53 unimplemented: type:"any"

noahdietz:
With #126 merged, I am going to close this and we can focus on the plugin work via https://github.com/googleapis/gnostic-go-generator/issues/7

PR #126, opened 2019-05-23
handle 'any' type

glickbot:
fix for #125 

Added 'any' type handling in service.proto, openapiv3 model and go plugin.

I'm where the best place to add tests for this would be, and I'm not sure if 'any' is going to work with openapiv2, etc (I've only tested openapiv3), but it's working with the bigquery v2 example (mentioned in the issue), so I wanted to get thoughts on the direction of this before I spent more time on it.


noahdietz:
@timburks can you take a look at this when you get a chance?

PR #128, opened 2019-06-10
Enhancements for the surface model

LorenzHW:
- Surface model now includes types from components/Parameters, components/RequestBodies and components/Responses sections from the OpenAPI spec
- Surface model now includes types that use the reference object ($ref) inside of OpenAPI spec for parameters, request bodies, responses

### DIFF
--- a/surface/model_openapiv3.go
+++ b/surface/model_openapiv3.go
@@ -51,9 +51,26 @@ func (b *OpenAPI3Builder) buildModel(document *openapiv3.Document) (*Model, erro
 
 // build builds an API service description, preprocessing its types and methods for code generation.
 func (b *OpenAPI3Builder) build(document *openapiv3.Document) (err error) {
+	err = b.buildTypesFromComponents(document.Components)
+	if err != nil {
+		return err
+	}
+
+	// Collect service method descriptions from each PathItem.
+	if document.Paths != nil {
+		for _, pair := range document.Paths.Path {
+			b.buildMethodFromPathItem(pair.Name, pair.Value)
+		}
+	}
+	return err
+}
+
+// buildTypesFromComponents builds multiple service type description from the "Components" section
+// in the OpenAPI specification.
+func (b *OpenAPI3Builder) buildTypesFromComponents(components *openapiv3.Components) (err error) {
 	// Collect service type descriptions from Components/Schemas.
-	if document.Components != nil && document.Components.Schemas != nil {
-		for _, pair := range document.Components.Schemas.AdditionalProperties {
+	if components != nil && components.Schemas != nil {
+		for _, pair := range components.Schemas.AdditionalProperties {
 			t, err := b.buildTypeFromSchemaOrReference(pair.Name, pair.Value)
 			if err != nil {
 				return err
@@ -63,12 +80,38 @@ func (b *OpenAPI3Builder) build(document *openapiv3.Document) (err error) {
 			}
 		}
 	}
-	// Collect service method descriptions from each PathItem.
-	if document.Paths != nil {
-		for _, pair := range document.Paths.Path {
-			b.buildMethodFromPathItem(pair.Name, pair.Value)
+	// Collect service type descriptions from Components/Parameters.
+	if components != nil && components.Parameters != nil {
+		for _, pair := range components.Parameters.AdditionalProperties {
+			parameters := []*openapiv3.ParameterOrReference{pair.Value}
+			_, err := b.buildTypeFromParameters(pair.Name, parameters, nil, true)
+			if err != nil {
+				return err
+			}
 		}
 	}
+	// Collect service type descriptions from Components/requestBodies
+	if components != nil && components.RequestBodies != nil {
+		for _, pair := range components.RequestBodies.AdditionalProperties {
+			t, err := b.buildTypeFromRequestBody(pair.Name, pair.Value, nil)
+
+			if err != nil {
+				return err
+			}
+			if t != nil {
+				b.model.addType(t)
+			}
+		}
+	}
+	// Collect service type descriptions from Components/responses
+	if components != nil && components.Responses != nil {
+		for _, pair := range components.Responses.AdditionalProperties {
+			namedResponseOrReference := []*openapiv3.NamedResponseOrReference{pair}
+			responses := &openapiv3.Responses{ResponseOrReference: namedResponseOrReference}
+			b.buildTypeFromResponses(pair.Name, responses, true)
+		}
+	}
+
 	return err
 }
 
@@ -144,8 +187,8 @@ func (b *OpenAPI3Builder) buildMethodFromPathItem(
 				m.Name = generateOperationName(method, path)
 			}
 			m.Description = op.Description
-			m.ParametersTypeName, err = b.buildTypeFromParameters(m.Name, op.Parameters, op.RequestBody)
-			m.ResponsesTypeName, err = b.buildTypeFromResponses(&m, m.Name, op.Responses)
+			m.ParametersTypeName, err = b.buildTypeFromParameters(m.Name, op.Parameters, op.RequestBody, false)
+			m.ResponsesTypeName, err = b.buildTypeFromResponses(m.Name, op.Responses, false)
 			b.model.addMethod(&m)
 		}
 	}
@@ -156,10 +199,15 @@ func (b *OpenAPI3Builder) buildMethodFromPathItem(
 func (b *OpenAPI3Builder) buildTypeFromParameters(
 	name string,
 	parameters []*openapiv3.ParameterOrReference,
-	requestBody *openapiv3.RequestBodyOrReference) (typeName string, err error) {
+	requestBody *openapiv3.RequestBodyOrReference,
+	fromComponent bool) (typeName string, err error) {
 	t := &Type{}
 	t.Name = name + "Parameters"
 	t.Description = t.Name + " holds parameters to " + name
+	if fromComponent {
+		t.Name = name
+		t.Description = t.Name + " is a parameter"
+	}
 	t.Kind = TypeKind_STRUCT
 	t.Fields = make([]*Field, 0)
 	for _, parametersItem := range parameters {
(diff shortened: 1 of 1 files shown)


surface/model_openapiv3.go
```diff
@@ -185,40 +233,75 @@ func (b *OpenAPI3Builder) buildTypeFromParameters(
...
+	return "", err
+}
+
+// buildTypeFromRequestBody builds a service type description from the request bodies of an OpenAPI
+// description. If tIn is not given, a new type is created. Otherwise tIn is used.
+func (b *OpenAPI3Builder) buildTypeFromRequestBody(name string, requestBody *openapiv3.RequestBodyOrReference, tIn *Type) (tOut *Type, err error) {
+	if tIn == nil {
+		tOut = &Type{}
+		tOut.Name = name
+	} else {
```
noahdietz: Personally, I don't like to use `else` statements if I can. They make the code a little harder to read, having to jump back up and through the `if`/`else if` statements to build the context in which the `else` body is being run.

I would do default initialization first, then do the check `tIn != nil` case. Like:

```go
tOut = &Type{
  Name: name,
}

if tIn != nil {
  tOut = tIn
}
```

This is easier to read: create a new Type with `name`, but if given one, use that instead.

surface/model_openapiv3.go
```diff
@@ -185,40 +233,75 @@ func (b *OpenAPI3Builder) buildTypeFromParameters(
...
+		tOut = tIn
+	}
+
 	if requestBody != nil {
 		content := requestBody.GetRequestBody().GetContent()
+		var f Field
```
noahdietz:
```go
f := Field{
  Position: Position_BODY,
  Serialize: true,
}
```

noahdietz: I think I see some other spots in the original code (not what is changed in this PR) that uses the style shown in this PR. So I understand following that as convention. But I believe initializing via a composite literal is the "Go way".

surface/model_openapiv3.go
```diff
@@ -282,10 +368,12 @@ func (b *OpenAPI3Builder) typeForSchema(schema *openapiv3.Schema) (kind FieldKin
 					a := items.GetSchemaOrReference()
 					if a[0].GetReference().GetXRef() != "" {
 						return FieldKind_ARRAY, typeForRef(a[0].GetReference().GetXRef()), format
-					} else if a[0].GetSchema().Type == "string" {
-						return FieldKind_ARRAY, "string", format
-					} else if a[0].GetSchema().Type == "object" {
-						return FieldKind_ARRAY, "object", format
+					} else {
+						// Determine whether the items of the array is a known type
+						types := map[string]bool{"string": true, "integer": true, "number": true, "boolean": true, "array": true, "object": true}
```
noahdietz: can we move `types` to a global variable? I'd prefer to avoid initializing a new map on every execution of this function when the contents doesn't ever change. Maybe give it a name like `knownTypes` too.

surface/model_openapiv3.go
```diff
@@ -282,10 +368,12 @@ func (b *OpenAPI3Builder) typeForSchema(schema *openapiv3.Schema) (kind FieldKin
...
+						if types[a[0].GetSchema().Type] {
```
noahdietz: When moving `types` out to a global variable, lets also make this part of the `if` chain as an `else if` please!

```go
if a[0].GetReference().GetXRef() != "" {
  ...
} else if types[a[0].GetSchema().Type] {
  ...
}
```

surface/model_openapiv3.go
```diff
@@ -185,40 +233,75 @@ func (b *OpenAPI3Builder) buildTypeFromParameters(
...
-	m *Method,
 	name string,
-	responses *openapiv3.Responses) (typeName string, err error) {
+	responses *openapiv3.Responses,
+	fromComponent bool) (typeName string, err error) {
 	t := &Type{}
 	t.Name = name + "Responses"
 	t.Description = t.Name + " holds responses of " + name
+
+	if fromComponent {
```
noahdietz: Same comment as above regarding where to put `if fromComponent`

surface/model_openapiv3.go
```diff
@@ -156,10 +199,15 @@ func (b *OpenAPI3Builder) buildMethodFromPathItem(
 func (b *OpenAPI3Builder) buildTypeFromParameters(
 	name string,
 	parameters []*openapiv3.ParameterOrReference,
-	requestBody *openapiv3.RequestBodyOrReference) (typeName string, err error) {
+	requestBody *openapiv3.RequestBodyOrReference,
+	fromComponent bool) (typeName string, err error) {
 	t := &Type{}
 	t.Name = name + "Parameters"
 	t.Description = t.Name + " holds parameters to " + name
+	if fromComponent {
```
noahdietz: can we put this code following the rest of the default initialization of `t`? Just to avoid breaking it up. `if fromComponent` is a "special" case and it reduces readability to drop that in the middle of default init of a struct

```go
t := &Type{}
t.Name = name + "Responses"
t.Description = t.Name + " holds responses of " + name
t.Kind = TypeKind_STRUCT
t.Fields = make([]*Field, 0)

if fromComponent { 
...
```

surface/model_openapiv3.go
```diff
@@ -185,40 +233,75 @@ func (b *OpenAPI3Builder) buildTypeFromParameters(
 			}
 			f.Serialize = true
 			t.addField(&f)
+		} else if parameterRef := parametersItem.GetReference(); parameterRef != nil {
+			f.Type = typeForRef(parameterRef.GetXRef())
+			f.Name = strings.ToLower(f.Type)
+			f.Serialize = false
```
noahdietz: This is unnecessary because the default value for a `bool` is `false`.

surface/model_openapiv3.go
```diff
@@ -51,9 +51,26 @@ func (b *OpenAPI3Builder) buildModel(document *openapiv3.Document) (*Model, erro
...
+	return err
+}
+
+// buildTypesFromComponents builds multiple service type description from the "Components" section
+// in the OpenAPI specification.
+func (b *OpenAPI3Builder) buildTypesFromComponents(components *openapiv3.Components) (err error) {
 	// Collect service type descriptions from Components/Schemas.
-	if document.Components != nil && document.Components.Schemas != nil {
-		for _, pair := range document.Components.Schemas.AdditionalProperties {
+	if components != nil && components.Schemas != nil {
```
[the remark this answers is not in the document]

surface/model_openapiv3.go

surface/model_openapiv3.go
LorenzHW: Done. I also changed it for the other spots in the original code to keep it more consistent

surface/model_openapiv3.go

surface/model_openapiv3.go

surface/model_openapiv3.go

surface/model_openapiv3.go

surface/model_openapiv3.go

surface/model_openapiv3.go
```diff
@@ -201,18 +209,21 @@ func (b *OpenAPI3Builder) buildTypeFromParameters(
 	parameters []*openapiv3.ParameterOrReference,
 	requestBody *openapiv3.RequestBodyOrReference,
 	fromComponent bool) (typeName string, err error) {
-	t := &Type{}
-	t.Name = name + "Parameters"
+	t := &Type{
```
noahdietz: I would change this to: 

```go
pName := name + "Parameters"
t := &Type{
  Name: pName,
  Kind: TypeKind_STRUCT,
  Fields: make([]*Field, 0),
  Description: pName + " holds parameters to " + name,
}
```

surface/model_openapiv3.go
```diff
@@ -185,54 +244,94 @@ func (b *OpenAPI3Builder) buildTypeFromParameters(
...
+		Name: name,
+	}
+
+	if tIn != nil {
+		tOut = tIn
+	}
+
 	if requestBody != nil {
 		content := requestBody.GetRequestBody().GetContent()
+		f := Field{
```
noahdietz: same nit here `f := &Field{`

surface/model_openapiv3.go
```diff
@@ -185,54 +244,94 @@ func (b *OpenAPI3Builder) buildTypeFromParameters(
...
 // buildTypeFromResponses builds a service type description from the responses of an API method
 func (b *OpenAPI3Builder) buildTypeFromResponses(
-	m *Method,
 	name string,
-	responses *openapiv3.Responses) (typeName string, err error) {
-	t := &Type{}
-	t.Name = name + "Responses"
+	responses *openapiv3.Responses,
+	fromComponent bool) (typeName string, err error) {
+	t := &Type{
```
noahdietz: same kind of nit as above regarding ordering:

```go
rName := name + "Responses"
t := &Type{
  Name: rName,
  ...
  Description: rName + " holds responses of " + name,
}
```

surface/model_openapiv3.go
```diff
@@ -68,8 +70,12 @@ func (b *OpenAPI3Builder) build(document *openapiv3.Document) (err error) {
 // buildTypesFromComponents builds multiple service type description from the "Components" section
 // in the OpenAPI specification.
 func (b *OpenAPI3Builder) buildTypesFromComponents(components *openapiv3.Components) (err error) {
+	if components == nil {
+		return err
```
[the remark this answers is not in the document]

surface/model_openapiv3.go
```diff
@@ -131,10 +137,11 @@ func (b *OpenAPI3Builder) buildTypeFromSchemaOrReference(
 			}
 			for _, pair2 := range schema.Properties.AdditionalProperties {
 				if schema := pair2.Value; schema != nil {
-					var f Field
-					f.Name = pair2.Name
+					f := Field{
```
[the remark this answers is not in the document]

surface/model_openapiv3.go
```diff
@@ -178,15 +185,16 @@ func (b *OpenAPI3Builder) buildMethodFromPathItem(
 			op = pathItem.Trace
 		}
 		if op != nil {
-			var m Method
-			m.Operation = op.OperationId
-			m.Path = path
-			m.Method = method
-			m.Name = sanitizeOperationName(op.OperationId)
+			m := Method{
```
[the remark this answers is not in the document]

surface/model_openapiv3.go

surface/model_openapiv3.go

surface/model_openapiv3.go

LorenzHW:
> A few more little nits for you, Lorenz! Thanks for your patience and willingness to work with me here :)

No worries about the nits Noah! It is only good for me :) Thanks for taking the time and reviewing the code. One question: Why do we prefer pointers over normal variables?

timburks:
Noah, whenever you think this is ready, let's squash-and-merge these changes in one commit.

noahdietz:
> Why do we prefer pointers over normal variables?

In this case, it was a style thing for me. We are only manipulating the data and then supplying it to a function that takes a pointer. A more specific reason might be that `Field` methods take a pointer receiver.

Generally, there is some guidance on [when to pass values](https://github.com/golang/go/wiki/CodeReviewComments#pass-values) and [what the receiver type should be](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type).

PR #130, opened 2019-06-22
Adding the position of a reference field.

LorenzHW:
#### Problem description:
Given following OpenAPI description: 

    paths:
      /testParameterReference:
        get:
          operationId: testParameterReference
          parameters:
            - $ref: '#/components/parameters/Parameter1'
          responses:
            200:
              description: success    
              
    components:
      parameters:
        Parameter1:
          name: param1
          in: query
          schema:
            type: integer
            format: int64
                  

The surface model creates a field with following properties:
Screenshot from 2019-06-22 14-15-22

Notice the Position property of the field that defaults to Position_BODY which is not correct.

#### Proposed solution
This PR sets the Position of the field to the referenced parameters position (in this case Position_QUERY).
An alternative solution would be to use a pointer for the 'Position' property and let it default to nil instead of Position_BODY.

Additionally, I am not sure about the [switch statement](https://github.com/LorenzHW/gnostic/blob/e6b1c2fad08b4c9c8abd19c877772ce33d0eeaef/surface/model_openapiv3.go#L231) above: according to the [OpenAPI spec](https://swagger.io/specification/#parameterObject)
the parameter object can only take: "header", "path", "query", or "cookie" for the 'in' field. I'd be happy to change that if necessary.

### DIFF
--- a/surface/model_openapiv3.go
+++ b/surface/model_openapiv3.go
@@ -407,8 +407,8 @@ func (b *OpenAPI3Builder) typeForSchema(schema *openapiv3.Schema) (kind FieldKin
 		}
 	}
 	if schema.AdditionalProperties != nil {
-		schemarOrReference := schema.AdditionalProperties.GetSchemaOrReference()
-		k, t, f := b.typeForSchemaOrReference(schemarOrReference)
+		schemaOrReference := schema.AdditionalProperties.GetSchemaOrReference()
+		k, t, f := b.typeForSchemaOrReference(schemaOrReference)
 		mapValueType := ""
 		if k == FieldKind_ARRAY {
 			mapValueType = "[]"
@@ -416,7 +416,7 @@ func (b *OpenAPI3Builder) typeForSchema(schema *openapiv3.Schema) (kind FieldKin
 		if f != "" {
 			t = f
 		}
-		mapValueType = mapValueType + t
+		mapValueType += t
 		return FieldKind_MAP, "map[string]" + mapValueType, ""
 	}
 	// this function is incomplete... use generic interface{} for now
--- a/surface/model_openapiv3.go
+++ b/surface/model_openapiv3.go
@@ -407,12 +407,17 @@ func (b *OpenAPI3Builder) typeForSchema(schema *openapiv3.Schema) (kind FieldKin
 		}
 	}
 	if schema.AdditionalProperties != nil {
-		additionalProperties := schema.AdditionalProperties
-		if propertySchema := additionalProperties.GetSchemaOrReference().GetReference(); propertySchema != nil {
-			if ref := propertySchema.XRef; ref != "" {
-				return FieldKind_MAP, "map[string]" + typeForRef(ref), ""
-			}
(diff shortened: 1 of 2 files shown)


LorenzHW:
The second commit handles following situation:

    paths:
      /testResponseMultipleContent:
        get:
          operationId: testResponseMultipleContent
          responses:
            200:
              description: success
              content:
                application/json:
                  schema:
                    $ref: '#/components/schemas/Person'
                application/xml:
                  schema:
                    $ref: '#/components/schemas/Person'
                    
                    
Note that 'content' contains application/json as well as application/xml. Hence, the response type
adds the same field twice (note the memory addresses on the screenshot (0xc4200ca480)).
Screenshot from 2019-06-24 08-47-36

surface/model_openapiv3.go
```diff
@@ -407,12 +407,17 @@ func (b *OpenAPI3Builder) typeForSchema(schema *openapiv3.Schema) (kind FieldKin
 		}
 	}
 	if schema.AdditionalProperties != nil {
-		additionalProperties := schema.AdditionalProperties
-		if propertySchema := additionalProperties.GetSchemaOrReference().GetReference(); propertySchema != nil {
-			if ref := propertySchema.XRef; ref != "" {
-				return FieldKind_MAP, "map[string]" + typeForRef(ref), ""
-			}
+		schemarOrReference := schema.AdditionalProperties.GetSchemaOrReference()
```
timburks: nit: could you change this to "schemaOrReference" (remove the extra 'r')?

surface/type.go
```diff
@@ -26,11 +26,20 @@ func (s *Type) FieldWithName(name string) *Field {
 	if s == nil || s.Fields == nil || name == "" {
 		return nil
 	}
+	// Compares Go specific field names.
```
timburks: With the change that you made below, I wonder if we should keep this test.

It would be nice if anything language-specific could be factored out of this layer.

surface/model_openapiv3.go
```diff
@@ -402,14 +407,30 @@ func (b *OpenAPI3Builder) typeForSchema(schema *openapiv3.Schema) (kind FieldKin
...
+		k, t, f := b.typeForSchemaOrReference(schemarOrReference)
+		mapValueType := ""
+		if k == FieldKind_ARRAY {
+			mapValueType = "[]"
+		}
+		if f != "" {
+			t = f
 		}
+		mapValueType = mapValueType + t
```
noahdietz: `mapValueType += t`

surface/type.go
LorenzHW: Yes, I was also thinking that. However, I checked gnostic-go-generator where this method is [used](https://github.com/googleapis/gnostic-go-generator/blob/ab235227442a8031d4a2dac629b992fa1c0605d7/render_server.go#L122) with language-specific values.

surface/model_openapiv3.go

surface/model_openapiv3.go

surface/type.go
timburks: OK, kudos for confirming that :) Let's go ahead like this and revisit the language-specific values later.

PR #132, opened 2019-07-05
Remove Swift code.

timburks:
I'd like to move all Swift code to a separate repo (probably to be named gnostic-swift), with its own CI so that all Swift work can stay tested without burdening contributions to the gnostic core.


noahdietz:
Do we have another repo set up? I'm happy to approve it, but we might want to wait on merging until the new home is up and can replace this. WDYT?

timburks:
swift code is in timburks/gnostic-swift

PR #135, opened 2019-07-26
Adding symbolic references to the surface model

LorenzHW:
Adds symbolic references to the surface model. A symbolic reference is an URL to another OpenAPI description inside of the input OpenAPI description.


plugins/environment.go
```diff
@@ -115,11 +116,17 @@ When the -plugin option is specified, these flags are ignored.`)
 		err = proto.Unmarshal(apiData, documentv3)
 		if err == nil {
 			env.Request.AddModel("openapi.v3.Document", documentv3)
+			sourceName := strings.ReplaceAll(*input, ".pb", ".yaml")
```
noahdietz: `strings.ReplaceAll` is a `go 1.12` API, it's causing CI to fail, which  is using `go 1.11`. It's reasonable to support `1.11` so let's use `strings.Replace(*input, ".pb", ".yaml", -1)`.

plugins/environment.go
```diff
@@ -115,11 +116,17 @@ When the -plugin option is specified, these flags are ignored.`)
...
+			if _, err := os.Stat(sourceName); os.IsNotExist(err) {
+				// sourceName does not exist exist. Lets try .json instead
+				sourceName = strings.ReplaceAll(*input, ".pb", ".json")
```
noahdietz: same comment regarding `strings.ReplaceAll`

plugins/environment.go
```diff
@@ -115,11 +116,17 @@ When the -plugin option is specified, these flags are ignored.`)
 		err = proto.Unmarshal(apiData, documentv3)
 		if err == nil {
 			env.Request.AddModel("openapi.v3.Document", documentv3)
+			sourceName := strings.ReplaceAll(*input, ".pb", ".yaml")
+			if _, err := os.Stat(sourceName); os.IsNotExist(err) {
+				// sourceName does not exist exist. Lets try .json instead
```
noahdietz: what if the `.json` version doesn't exist? I'm assuming it will fail elsewhere? Might be worth it to fail here.

plugins/environment.go
LorenzHW: Done

plugins/environment.go
LorenzHW: Done

plugins/environment.go
LorenzHW: Done

LorenzHW:
Commit 6f18f45 correctly constructs the filenames now. I didn't realize that there could also be a reference to a local file. Basically, I reverted what I did in commit 895ee40 (first commit) and added an additional check whether we have an URL.

plugins/environment.go
```diff
@@ -115,11 +116,21 @@ When the -plugin option is specified, these flags are ignored.`)
 		err = proto.Unmarshal(apiData, documentv3)
 		if err == nil {
 			env.Request.AddModel("openapi.v3.Document", documentv3)
+			sourceName := strings.Replace(*input, ".pb", ".yaml", -1)
+			if _, err := os.Stat(sourceName); os.IsNotExist(err) {
+				// sourceName does not exist. Lets try .json instead
+				sourceName = strings.Replace(*input, ".pb", ".json", -1)
```
timburks: I'm not clear on how this code is getting exercised. Referring back to line 37/38, it says that we expect the input to be in binary form. Are we now also allowing JSON/YAML input?

surface/model_openapiv3.go
```diff
@@ -39,11 +40,12 @@ func newOpenAPI3Builder() *OpenAPI3Builder {
 	return &OpenAPI3Builder{model: &Model{}}
 }
 
-func (b *OpenAPI3Builder) buildModel(document *openapiv3.Document) (*Model, error) {
+func (b *OpenAPI3Builder) buildModel(document *openapiv3.Document, sourceName string) (*Model, error) {
```
timburks: I notice that we're only adding symbolic references for OpenAPI v3. Perhaps we should note that explicitly.

compiler/reader.go
```diff
@@ -174,10 +175,16 @@ func ReadInfoForRef(basefile string, ref string) (interface{}, error) {
...
+	basedir, _ := filepath.Split(basefile)
 	parts := strings.Split(ref, "#")
 	var filename string
 	if parts[0] != "" {
-		filename = parts[0]
+		if _, err := url.ParseRequestURI(parts[0]); err != nil {
+			// It is not an URL, so the file is local
+			filename = basedir + parts[0]
+		} else {
+			filename = parts[0]
```
noahdietz: let's make this the default behavior and handle local file as a special case.

```go
filename = parts[0]
if _, err := url.ParseRequestURI(parts[0]); err != nil {
    // It is not an URL, so the file is local
    filename = basedir + parts[0]
}
```

This is code flows straight down and thus slightly more readable

plugins/environment.go
LorenzHW: We still expect the input to be in binary form. This is the same sourceName as in the [gnostic.go](https://github.com/googleapis/gnostic/pull/135/files#diff-07b5e04f760e7b7bac224fec4ccf7f7cR158) file. However, in the case above we run the plugin as command line tool with a binary file as input. So our best bet is to guess the sourceName from which the binary file is coming from.

compiler/reader.go
LorenzHW: Ahhh! Done. Old habits. Thanks Noah.

surface/model_openapiv3.go
LorenzHW: Sure, I can add a comment. Where is the best place to add the comment? I believe inside surface.proto right next to the `symbolicReferences` field it would be likely to get spotted.

surface/surface.proto
```diff
@@ -88,4 +88,6 @@ message Model {
     string name = 1; // a free-form title for the API
     repeated Type types = 2; // the types used by the API
     repeated Method methods = 3; // the methods (functions) of the API
+    repeated string symbolicReferences = 4; // references to other OpenAPI files
+    string sourceName = 5; // source name of the OpenAPI description (.yaml or .json)
```
noahdietz: please use snake_case for such compound names (will probably need to regenerate the Go file, but it shouldn't break your code).

Sorry I missed this earlier.

LorenzHW: Done. Other fields inside surface.proto are still cameCase. Should I change them as well?

noahdietz: Eh no, that's ok. I just prefer not to continue to add to it. But that's just me. Even if we did, it would be in a different PR.

LorenzHW:
Deprecated in favor of #136.

PR #136, opened 2019-08-04
New surface model

LorenzHW:
As I was optimizing gnostic-grpc with the OpenAPI descriptions from [APIs-guru/openapi-directory](https://github.com/APIs-guru/openapi-directory), I realized that the current implementation of the surface model is not properly handling nested schemas and other cases.

This implementation builds the types and methods of the surface model based on recursion and therefore handles nested schemas. Additionally it includes changes from #135 .

I haven't tested this with [gnostic-go-generator](https://github.com/googleapis/gnostic-go-generator) but in theory it should construct a richer output for every plugin. However, surface model Types are not exactly constructed the same way as they were with the previous implementation so initial test cases might fail.


timburks:
Does this replace #135 or should we merge both?

LorenzHW:
This replaces #135.

timburks:
Merged. Thank you!

PR #138, opened 2019-08-14
Remove generated pb code.

timburks:
Now all protobuf support code is generated when the project is built (using the Makefile).


noahdietz:
Might I suggest we also add the `*.pb.go` files to the `.gitignore` so this doesn't get annoying?

timburks:
@noahdietz thanks for suggesting that! Will do.

PR #141, opened 2019-08-15
Re-adding generated files to avoid problems with "go get".

timburks:
https://github.com/googleapis/gnostic/issues/139


niting3c:
the version 0.5.5 is still having the same issue
I downgraded to 0.5.0 , it works there fine

PR #149, opened 2019-09-20
Updates to OpenAPI v2 schema.

Capstan:
From last commit at https://github.com/OAI/OpenAPI-Specification/commit/88cd94419e117b154b67b834fa8e471bb98bd346.


timburks:
Thanks for suggesting this! The comment fix (line 206) was made separately and since the other changes would change the proto model (and I don't think they change the set of descriptions that we can read), we're going to omit this change.

PR #153, opened 2020-01-25
Reorganize tests.

timburks:
The top-level tests now use a "lib" version of Gnostic
to run in a single process, which allows them to run in
test environments that require that.

Plugin and extension tests are moved into their respective
directories and are run from "make test" which is also
used by .travis.yaml to run these tests in CI. One sample
app (petstore-builder) is also tested in this way.

To avoid confusing the "make" tool, the "test" directory
is renamed "testdata".


Makefile
```diff
@@ -1,7 +1,12 @@
...
 	./COMPILE-PROTOS.sh
 	go get ./...
 	go install ./...
 	cd extensions/sample; make
 
+test:
+	go test . -v
+	cd plugins; go test . -v
+	cd extensions; go test . -v
+	cd apps/petstore-builder; go test . -v
```
noahdietz: This can be simplified with `go test ./... -v`. Unless the goal was to only test specific packages, in which case, `cd` is unnecessary, you can provide the relative path of a package to test:

```
go test . -v
go test ./plugins -v
go test ./extensions -v
go test ./apps/petstore-builder -v
```

PR #154, opened 2020-01-30
Update Makefile to run all tests (and fixed broken tests).

timburks:
Tests in compiler and generate-gnostic were broken. This fixes them and adds them to CI testing.

### DIFF
--- a/apps/disco/main.go
+++ b/apps/disco/main.go
@@ -44,7 +44,8 @@ Usage:
 	if arguments["help"].(bool) {
 		fmt.Println("\nRead and process Google's Discovery Format for APIs.")
 		fmt.Println(usage)
-		fmt.Println("To learn more about Discovery Format, visit https://developers.google.com/discovery/\n")
+		fmt.Println("To learn more about Discovery Format, visit https://developers.google.com/discovery/.")
+		fmt.Println()
 	}
 
 	// List APIs.
--- a/compiler/reader_test.go
+++ b/compiler/reader_test.go
@@ -15,7 +15,7 @@
 package compiler
 
 import (
-	. "gopkg.in/check.v1"
+	"gopkg.in/check.v1"
 	"io"
 	"net/http"
 	"testing"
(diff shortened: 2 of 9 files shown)


apps/disco/main.go
```diff
@@ -44,7 +44,8 @@ Usage:
 	if arguments["help"].(bool) {
 		fmt.Println("\nRead and process Google's Discovery Format for APIs.")
 		fmt.Println(usage)
-		fmt.Println("To learn more about Discovery Format, visit https://developers.google.com/discovery/\n")
+		fmt.Println("To learn more about Discovery Format, visit https://developers.google.com/discovery/.")
+		fmt.Println()
```
timburks: This was to fix a compiler warning about an "extraneous newline" at the end of the string.

generate-gnostic/generate-extension_test.go
```diff
@@ -11,7 +11,7 @@ func TestErrorExtensionGeneratorUnsupportedPrimitive(t *testing.T) {
 	var err error
 
 	output, err := exec.Command(
-		"generator",
+		"generate-gnostic",
 		"--extension",
```
timburks: The generator was being run with the wrong (an old) name.

compiler/reader_test.go
```diff
@@ -71,6 +71,7 @@ func (s *ReaderTestingSuite) TestDisableInfoCache(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(petstore, NotNil)
 	c.Assert(len(infoCache), Equals, 0)
+	EnableInfoCache()
```
timburks: Caches are disabled at the start of this (and the following) test. In an order-dependence, subsequent tests failed because they expected the caches to be enabled.

generate-gnostic/test/errors/x-extension-name-collision.errors
```diff
@@ -1,2 +1 @@
-Schema SampleCompanyTwoPrimitiveString and SampleCompanyOnePrimitiveString have the same 'id' field value.
-
+schema SampleCompanyTwoPrimitiveString and SampleCompanyOnePrimitiveString have the same 'id' field value
```
timburks: Go guidance is for error messages to be uncapitalized and *not* end with a period.

PR #155, opened 2020-01-31
Lower-case OpenAPIv2 and OpenAPIv3 dirs to follow style guide.

timburks:
https://blog.golang.org/package-names


timburks:
Motivation: we noticed that one of these directories was being forcibly-lowercased when it was imported for internal use. Checking the style guide, we found they were non-compliant, so it seemed best to change their names.

VineethReddy02:
@timburks making changes around renaming packages is breaking other dependent projects for example 
https://github.com/kubernetes/client-go/issues/743
IMO Maybe we should avoid such changes.

timburks:
Hi @VineethReddy02 - that's unfortunate! My understanding was that the Kubernetes project used a tagged version of gnostic - and it seems that there's a tagged reference in your [go.mod](https://github.com/kubernetes/client-go/blob/21de178e1daf31a33f605baff694b7b95c943311/go.mod#L19). Were you not building with modules?

VineethReddy02:
@timburks I was using gitmodules, So, unfortunately, it tries to pull gnostic latest. I fixed this issue by checking out gnostic to the previous version.
Thanks for the update!

boyan-soubachov:
Had the exact same issue, to solve (temporarily), pin to v0.4.0 by doing the following in your go-modules-enabled project:

`go get github.com/googleapis/gnostic@v0.4.0`

I have submitted a PR to `github.com/kubernetes/kubernetes` to use the newly-lowercased `openapiv2` dependency.

PR #158, opened 2020-03-08
Add mtlsRootUrl property to Discovery schema.

timburks:
Add a missing (new) property to the Discovery schema.


timburks:
This new property was causing the discovery reader to fail to load schemas.

(Currently the reader is very strict about rejecting unknown properties - this is worth reconsidering)

PR #163, opened 2020-04-21
Synchronize accesses to reader cache maps.

timburks:
While writing a couple of tools that mass-process API specs in parallel goroutines, I found some synchronization problems caused by the shared caches used in the reader. This PR protects all of the public functions in compiler/reader.go with mutexes for the file and info caches.

### DIFF
--- a/compiler/reader.go
+++ b/compiler/reader.go
@@ -35,6 +35,14 @@ var infoCache map[string]interface{}
 var fileCacheEnable = true
 var infoCacheEnable = true
 
+// These locks are used to synchronize accesses to the fileCache and infoCache
+// maps (above). They are global state and can throw thread-related errors
+// when modified from separate goroutines. The general strategy is to protect
+// all public functions in this file with mutex Lock() calls. As a result, to
+// avoid deadlock, these public functions should not call other public
+// functions, so some public functions have private equivalents.
+// In the future, we might consider replacing the maps with sync.Map and
+// eliminating these mutexes.
 var fileCacheMutex sync.Mutex
 var infoCacheMutex sync.Mutex
 
--- a/compiler/reader.go
+++ b/compiler/reader.go
@@ -15,7 +15,6 @@
 package compiler
 
 import (
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -23,17 +22,21 @@ import (
 	"net/url"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	yaml "gopkg.in/yaml.v2"
 )
 
+var verboseReader = false
+
 var fileCache map[string][]byte
 var infoCache map[string]interface{}
-var count int64
 
-var verboseReader = false
 var fileCacheEnable = true
 var infoCacheEnable = true
+
+var fileCacheMutex sync.Mutex
+var infoCacheMutex sync.Mutex
 
 func initializeFileCache() {
 	if fileCache == nil {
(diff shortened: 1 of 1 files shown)


compiler/reader.go
```diff
@@ -47,60 +50,94 @@ func initializeInfoCache() {
...
 		return
 	}
 	initializeInfoCache()
 	delete(infoCache, filename)
 }
 
+// GetInfoCache returns the info cache map.
 func GetInfoCache() map[string]interface{} {
+	infoCacheMutex.Lock()
+	defer infoCacheMutex.Unlock()
```
noahdietz, line 106: Is the locking only necessary when `initializeInfoCache()` is called? Or is it also necessary to return the reference?

If the former, please move the locking/unlocking inside the conditional.

compiler/reader.go
```diff
@@ -47,60 +50,94 @@ func initializeInfoCache() {
...
 // FetchFile gets a specified file from the local filesystem or a remote location.
 func FetchFile(fileurl string) ([]byte, error) {
+	fileCacheMutex.Lock()
+	defer fileCacheMutex.Unlock()
+	return fetchFile(fileurl)
+}
+
+func fetchFile(fileurl string) ([]byte, error) {
 	var bytes []byte
 	initializeFileCache()
```
noahdietz, line 142: I know this isn't part of the diff, but should this `initializeFileCache()` call be synchronized? Or should the `initializeFileCache` function itself have lock calls in it? It _can_ mutate a global variable, so should it be synchronized?

compiler/reader.go
```diff
@@ -121,7 +158,7 @@ func FetchFile(fileurl string) ([]byte, error) {
 	}
 	defer response.Body.Close()
 	if response.StatusCode != 200 {
-		return nil, errors.New(fmt.Sprintf("Error downloading %s: %s", fileurl, response.Status))
+		return nil, fmt.Errorf("Error downloading %s: %s", fileurl, response.Status)
 	}
 	bytes, err = ioutil.ReadAll(response.Body)
 	if fileCacheEnable && err == nil {
```
noahdietz, line 164: Should the following assignment of `fileCache[fileurl] = bytes` be synchronized?

compiler/reader.go
timburks: I think we should protect the test of infoCache, but generally should clarify that the locking strategy here is to make all of the public functions in this directory exclusive. So every capitalized function begins with a lock/defer unlock and these functions never call other capitalized functions - accesses inside the uncapitalized (private) functions are unsynchronized because they are synchronized at the higher level of the public function interfaces.

timburks: I'll add a comment to clarify this in the code.

compiler/reader.go
timburks: This call is unsynchronized because it is in a private function (see my comment at line 106 of the new version)

compiler/reader.go
timburks: This is unsychronized because it is in a private function.

compiler/reader.go
```diff
@@ -15,26 +15,37 @@
...
 
+// These locks are used to synchronize accesses to the fileCache and infoCache
+// maps (above). They are global state and can throw thread-related errors
+// when modified from separate goroutines. The general strategy is to protect
+// all public functions in this file with mutex Lock() calls. As a result, to
+// avoid deadlock, these public functions should not call other public
+// functions, so some public functions have private equivalents.
+// In the future, we might consider replacing the maps with sync.Map and
+// eliminating these mutexes.
+var fileCacheMutex sync.Mutex
```
timburks, line 46: @noahdietz FYI

PR #166, opened 2020-04-25
Schema updates

timburks:
This contains updates and improvements to schema-based parsing.

The biggest change here is that the models for JSON references are now "open" and allow other fields to be specified alongside "$ref". Following the spec, these are ignored, where previously
they triggered error messages that caused some API descriptions to fail to parse.

This commit also contains small tweaks to the openapi-3.0.json schema that improves model coverage and a small refinement to error printing (usage information is only printed for usage-related errors, fixing a problem introduced in pr#161)


lib/gnostic.go
```diff
@@ -37,6 +37,21 @@ import (
...
 )
 
+// UsageError is a response to invalid command-line inputs
+// and should trigger display of usage (help) information.
+type UsageError struct {
+	Message string
+}
+
+func (e *UsageError) Error() string {
+	return fmt.Sprintf("%s", e.Message)
```
noahdietz, line 47: why do `fmt.Sprintf` here? Can it be just `return e .Message`?

lib/gnostic.go
```diff
@@ -37,6 +37,21 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+// UsageError is a response to invalid command-line inputs
+// and should trigger display of usage (help) information.
+type UsageError struct {
+	Message string
```
noahdietz, line 43: Is exporting `Message` necessary? Should `Error()` be the only way of accessing the error string, or is direct `Message` access also acceptable?

openapiv3/OpenAPIv3.go
```diff
@@ -2483,6 +2416,44 @@ func NewNamedString(in interface{}, context *compiler.Context) (*NamedString, er
...
 }
 
+// NewNamedStringArray creates an object of type NamedStringArray if possible, returning an error if not.
+func NewNamedStringArray(in interface{}, context *compiler.Context) (*NamedStringArray, error) {
+	errors := make([]error, 0)
+	x := &NamedStringArray{}
+	m, ok := compiler.UnpackMap(in)
+	if !ok {
+		message := fmt.Sprintf("has unexpected value: %+v (%T)", in, in)
+		errors = append(errors, compiler.NewError(context, message))
```
noahdietz, line 2426: Suggestion: put a `return` in this conditional block. `return x, compiler.NewErrorGroupOrNil(errors)`.

This keeps the error flow indented but the following "green path" code path in the `else` at minimal indentation.

Context: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

openapiv3/OpenAPIv3.go
```diff
@@ -4373,12 +4337,22 @@ func NewSecurityRequirement(in interface{}, context *compiler.Context) (*Securit
...
-		invalidKeys := compiler.InvalidKeysInMap(m, allowedKeys, allowedPatterns)
-		if len(invalidKeys) > 0 {
-			message := fmt.Sprintf("has invalid %s: %+v", compiler.PluralProperties(len(invalidKeys)), strings.Join(invalidKeys, ", "))
-			errors = append(errors, compiler.NewError(context, message))
+		// repeated NamedStringArray additional_properties = 1;
+		// MAP: StringArray
+		x.AdditionalProperties = make([]*NamedStringArray, 0)
+		for _, item := range m {
+			k, ok := compiler.StringValue(item.Key)
+			if ok {
```
noahdietz, line 4345: Suggestion: put a `continue` in a `!ok` conditional rather than have the "green path" code be indented.

```go
if !ok {
  continue
}

v := item.Value
// ...
```

lib/gnostic.go
timburks: yes, fixed.

lib/gnostic.go
timburks: Thanks, I'm making this private.

openapiv3/OpenAPIv3.go
timburks: this is generated code, I'd like to address it separately in issue #167

openapiv3/OpenAPIv3.go
timburks: as above, since this is generated code, I'd like to address it separately in issue #167

PR #169, opened 2020-05-09
Handle OneOf, AnyOf, and AllOf sub schemas

LorenzHW:
- This PR creates better types and fields in case a schema has sub schemas which are defined with OneOf, AnyOf, or AllOf relation. The first commit fixes [issue 22](https://github.com/googleapis/gnostic-grpc/issues/22) in gnostic-grpc.

For example `input.yaml`:
```
components:
  schemas:
    Pet:
      type: object
      required:
        - pet_type
      properties:
        pet_type:
          type: string
      discriminator:
        propertyName: pet_type
    Dog:     # "Dog" is a value for the pet_type property (the discriminator value)
      allOf: # Combines the main `Pet` schema with `Dog`-specific properties
        - $ref: '#/components/schemas/Pet'
        - type: object
          # all other properties specific to a `Dog`
          properties:
            bark:
              type: boolean
            breed:
              type: string
              enum: [Dingo, Husky, Retriever, Shepherd]
```
gnostic-grpc would now generate:
```
message Pet {
  string pet_type = 1;
}

message Dog {
  Pet pet = 1;

  bool bark = 2;

  string breed = 3;
}
```

As of now it generates:
```
message Pet {
  string pet_type = 1;
}

message DogAllOf2 {
  bool bark = 1;

  string breed = 2;
}

message Dog {
  Pet all_of_1 = 1;

  DogAllOf2 all_of_2 = 2;
}
```

- Second commit fixes [issue 21](https://github.com/googleapis/gnostic-grpc/issues/21) in gnostic-grpc.

### DIFF
--- a/surface/model_openapiv3.go
+++ b/surface/model_openapiv3.go
@@ -18,20 +18,20 @@ import (
 	"github.com/googleapis/gnostic/compiler"
 	openapiv3 "github.com/googleapis/gnostic/openapiv3"
 	"log"
-	"strings"
 )
 
 type OpenAPI3Builder struct {
-	model *Model
+	model    *Model
+	document *openapiv3.Document
 }
 
 // NewModelFromOpenAPIv3 builds a model of an API service for use in code generation.
 func NewModelFromOpenAPI3(document *openapiv3.Document, sourceName string) (*Model, error) {
-	return newOpenAPI3Builder().buildModel(document, sourceName)
+	return newOpenAPI3Builder(document).buildModel(document, sourceName)
 }
 
-func newOpenAPI3Builder() *OpenAPI3Builder {
-	return &OpenAPI3Builder{model: &Model{}}
+func newOpenAPI3Builder(document *openapiv3.Document) *OpenAPI3Builder {
+	return &OpenAPI3Builder{model: &Model{}, document: document}
 }
 
 // Fills the surface model with information from a parsed OpenAPI description. The surface model provides that information
@@ -39,7 +39,7 @@ func newOpenAPI3Builder() *OpenAPI3Builder {
 // Since OpenAPI schemas can be indefinitely nested, it is a recursive approach to build all Types and Methods.
 // The basic idea is that whenever we have "named OpenAPI object" (e.g.: NamedSchemaOrReference, NamedMediaType) we:
 //	1. Create a Type with that name
-//	2. Recursively execute according methods on child schemas (see buildFromSchema function)
+//	2. Recursively create sub schemas (see buildFromSchema function)
 // 	3. Return a FieldInfo object that describes how the created Type should be represented inside another Type as field.
 func (b *OpenAPI3Builder) buildModel(document *openapiv3.Document, sourceName string) (*Model, error) {
 	b.model.Types = make([]*Type, 0)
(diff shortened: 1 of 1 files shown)


surface/model_openapiv3.go
```diff
@@ -439,3 +436,34 @@ func (b *OpenAPI3Builder) buildFromSchema(name string, schema *openapiv3.Schema)
 	log.Printf("Unimplemented: could not find field info for schema: %v", schema)
 	return nil
 }
+
+// buildFromOneOfAnyOfAndAllOf adds appropriate fields to the 'schemaType' given a new 'schemaOrRef'.
+func (b *OpenAPI3Builder) buildFromOneOfAnyOfAndAllOf(schemaOrRef *openapiv3.SchemaOrReference, schemaType *Type) {
+	// Related: https://github.com/googleapis/gnostic-grpc/issues/22
+	if schema := schemaOrRef.GetSchema(); schema != nil {
+		fieldInfo := b.buildFromSchemaOrReference("ATemporaryTypeThatWillBeRemoved", schemaOrRef)
+		if t := findType(b.model.Types, "ATemporaryTypeThatWillBeRemoved"); t != nil {
```
timburks, line 445: Could you add a comment with some explanation of what is happening here? Is "ATemporaryType..." being used to collect fields from the OneOf/AnyOf/AllOfs and then "WillBeRemoved" when these fields are merged into the main type?

LorenzHW:
@timburks yes, you are correct. Basically we want to flatten every referenced sub schema:
Screen Shot 2020-05-10 at 18 47 43

surface/model_openapiv3.go
LorenzHW: Yes, exactly.

surface/model_openapiv3.go
```diff
@@ -189,11 +163,11 @@ func (b *OpenAPI3Builder) buildFromNamedPath(name string, pathItem *openapiv3.Pa
 	}
 }
 
-// Builds the "Parameters" and "Responses" types for an operation, adds them to the model, and returns the names of the types.
+// Builds the "Request" and "Responses" types for an operation, adds them to the model, and returns the names of the types.
 // If no such Type is added to the model an empty string is returned.
 func (b *OpenAPI3Builder) buildFromNamedOperation(name string, operation *openapiv3.Operation) (parametersTypeName string, responseTypeName string) {
 	// At first, we build the operations input parameters. This includes parameters (like PATH or QUERY parameters) and a request body
-	operationParameters := makeType(name + "Parameters")
+	operationParameters := makeType(name + "Request")
```
noahdietz, line 170: Are we sure we want to change the name here? "Parameters" is the OpenAPI schema property name, so I'd say it has some contextual value. I like "Request" because it lines up with protobuf style, but I'm not sure we want to do that here. 

Also, I don't quite remember, is this name used in a plugin/user-facing API? Meaning, will a downstream consumer (a plugin, an end-user of gnostic, etc) be depending on the `Name` being `FooParameters` (i.e. they are generating code with that type name)? If so, changing the constructed name could cause problem for downstream tools/users.

surface/model_openapiv3.go
```diff
@@ -439,3 +404,58 @@ func (b *OpenAPI3Builder) buildFromSchema(name string, schema *openapiv3.Schema)
...
+		} else {
+			// schemaOrRef is some kind of primitive schema (e.g. of type string)
+			makeFieldAndAppendToType(fieldInfo, schemaType, "value")
+		}
+	} else if ref := schemaOrRef.GetReference(); ref != nil {
+		referencedSchemaName := validTypeForRef(ref.XRef)
+		// Make sure that the referenced type exists, before we add the fields to the current schema
+		for _, namedSchema := range b.document.GetComponents().GetSchemas().GetAdditionalProperties() {
+			if referencedSchemaName == namedSchema.Name {
+				b.buildFromSchemaOrReference(namedSchema.Name, namedSchema.Value)
```
noahdietz, line 429: If we find the `referencedSchemaName`, should this loop `break`?

surface/model_openapiv3.go
```diff
@@ -439,3 +404,58 @@ func (b *OpenAPI3Builder) buildFromSchema(name string, schema *openapiv3.Schema)
...
+				schemaType.Fields = append(schemaType.Fields, f)
+			}
+			b.removeType(t)
+		} else {
+			// schemaOrRef is some kind of primitive schema (e.g. of type string)
+			makeFieldAndAppendToType(fieldInfo, schemaType, "value")
+		}
+	} else if ref := schemaOrRef.GetReference(); ref != nil {
+		referencedSchemaName := validTypeForRef(ref.XRef)
+		// Make sure that the referenced type exists, before we add the fields to the current schema
```
noahdietz, line 426: What happens if we _don't_ find it? Is that a possibility?

surface/model_openapiv3.go
```diff
@@ -439,3 +404,58 @@ func (b *OpenAPI3Builder) buildFromSchema(name string, schema *openapiv3.Schema)
...
+		for _, namedSchema := range b.document.GetComponents().GetSchemas().GetAdditionalProperties() {
+			if referencedSchemaName == namedSchema.Name {
+				b.buildFromSchemaOrReference(namedSchema.Name, namedSchema.Value)
+			}
+		}
+
+		if t := findType(b.model.Types, referencedSchemaName); t != nil {
+			for _, f := range t.Fields {
+				schemaType.Fields = append(schemaType.Fields, f)
+			}
```
noahdietz, line 436: Can we remove the loop and just do `schemaType.Fields = append(schemaType.Fields, t.Fields...)`?

surface/model_openapiv3.go
```diff
@@ -439,3 +404,58 @@ func (b *OpenAPI3Builder) buildFromSchema(name string, schema *openapiv3.Schema)
...
+func (b *OpenAPI3Builder) buildFromOneOfAnyOfAndAllOf(schemaOrRef *openapiv3.SchemaOrReference, schemaType *Type) {
+	// Related: https://github.com/googleapis/gnostic-grpc/issues/22
+	if schema := schemaOrRef.GetSchema(); schema != nil {
+		// Build a temporary type that has the required fields; add the fields to the current schema; remove the
+		// current type
+		fieldInfo := b.buildFromSchemaOrReference("ATemporaryTypeThatWillBeRemoved", schemaOrRef)
+		if t := findType(b.model.Types, "ATemporaryTypeThatWillBeRemoved"); t != nil {
+			for _, f := range t.Fields {
+				schemaType.Fields = append(schemaType.Fields, f)
+			}
```
noahdietz, line 418: Can we remove the loop and just do `schemaType.Fields = append(schemaType.Fields, t.Fields...)`?

surface/model_openapiv3.go
```diff
@@ -439,3 +404,58 @@ func (b *OpenAPI3Builder) buildFromSchema(name string, schema *openapiv3.Schema)
...
+			b.removeType(t)
+		} else {
+			// schemaOrRef is some kind of primitive schema (e.g. of type string)
+			makeFieldAndAppendToType(fieldInfo, schemaType, "value")
+		}
```
noahdietz, line 423: Style comment: could we invert the structuring of this `if`/`else` to reduce indentation? In the spirit of https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

```go
t := findType(b.model.Types, "ATemporaryTypeThatWillBeRemoved")
if t == nil {
    // schemaOrRef is some kind of primitive schema (e.g. of type string)
    makeFieldAndAppendToType(fieldInfo, schemaType, "value")
    return
}

schemaType.Fields = append(schemaType.Fields, t.Fields...)
b.removeType(t)
```

This puts the "primitive" scenario in the conditional flow with a terminating `return` and the "non-primitive" flow as the regular path of execution (in the scope of the `schema != nil` condition)

Feel free to ignore this, I just find it easier to read/cleaner.

surface/model_openapiv3.go
LorenzHW: Yes, I agree with you Noah. We should change the suffix in gnostic-grpc.

I think gnostic-go-generator would have generated something like:
```
type FooRequest {
   ...
}
```
Instead of:
```
type FooParameters {
   ...
}
```

surface/model_openapiv3.go

surface/model_openapiv3.go

surface/model_openapiv3.go

surface/model_openapiv3.go
LorenzHW: If the referenced schema is in another API description it would break.

Additionally, I am wondering whether we should do the same loop for:
```
b.document.GetComponents().GetResponses().GetAdditionalProperties()
```
and:
```
b.document.GetComponents().GetRequestBodies().GetAdditionalProperties()
```

surface/model_openapiv3.go
LorenzHW: Done!

🙈Thanks Noah. I think we had that two or three times already. When I wrote that if-else statement I was like: "Yup, Noah won't approve this"

surface/model_openapiv3.go
noahdietz: @timburks WDYT about the above?

surface/model_openapiv3.go
noahdietz: Ok, yeah good call to revert the change then. gnostic-grpc seems like the right spot for such a change 😄

surface/model_openapiv3.go
noahdietz: Hahaha thanks for fixing it. I'm relentless on the if-else statements 😉

surface/model_openapiv3.go
timburks: I don't think we had seriously considered OneOf/AnyOf/AllOfs before this, so this is progress (even if it breaks for some cases). Here it looks like the model will be incomplete if the ref can't be resolved. Perhaps we could warn about that, but otherwise, I'm ok with addressing that situation later.

LorenzHW: Done

timburks:
Thanks!

PR #171, opened 2020-05-15
Added fuzzer

AdamKorcz:
This PR adds a fuzzer that targes the gnostic `Main()` function. 

I have setup an application to integrate gnostic on oss-fuzz which can be found [here](https://github.com/google/oss-fuzz/pull/3822). 

Integration into oss-fuzz requires at least one contact email address for bug reports. I have put my own as [an example](https://github.com/google/oss-fuzz/pull/3822/files#diff-e5c753c016c087ba782f86620012d2d6R2).

If there is interest in integrating with oss-fuzz, I ask kindly for the contact email addresses to add to the list of receivers for bug reports.

### DIFF
--- a/lib/fuzz/fuzz.go
+++ b/lib/fuzz/fuzz.go
@@ -1,0 +1,30 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


timburks:
@AdamKorcz would it be worthwhile to merge this without your other PR to oss-fuzz? I'm wondering if the entire fuzzing setup (minus the tool itself) could be put in this repo.

AdamKorcz:
Yeah definitely. The fuzzer can be run by itself locally. See https://github.com/dvyukov/go-fuzz for info

PR #175, opened 2020-05-27
Vocabulary: New App

gizzon:
The purpose of this PR is to incorporate an additional app that will be used for obtaining name frequencies. This is done by including an additional 3 files which are responsible for collecting the  field names, schema names, operation names, and property names. Data structures generated from protocol buffers are then filled with the names and serialized.


apps/vocabulary/main.go
```diff
@@ -0,0 +1,261 @@
...
+	_, ok := mapName[word]
+	if ok {
+		mapName[word] += 1
+	} else {
+		mapName[word] = 1
+	}
+}
+
+func createCSV(names map[string]int) {
+	f, ferror := os.Create("IndexFreqSort.csv")
```
timburks, line 55: Since we've stopped calling this, let's take it out before committing (it might be worth hanging onto this code for use in a future helper app)

apps/vocabulary/main.go
```diff
@@ -0,0 +1,261 @@
...
+		return
+	}
+	for k, _ := range names {
+		temp := fmt.Sprintf("%s\n", k)
+		f4.WriteString(temp)
+	}
+	f4.Close()
+}
+
+func printDocument(document *pb.Document, schemas map[string]int, operationId map[string]int, names map[string]int, properties map[string]int) {
```
timburks, line 130: Could we change all instances of print[Something] to process[Something]? That would more accurately represent current usage (noting that these functions were derived from code in apps/report/main.go).

apps/vocabulary/vocabulary.proto
```diff
@@ -0,0 +1,28 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+syntax = "proto3";
+
```
timburks, line 17: Please insert a package declaration here. I think it should be this:

```package gnostic.metrics.v1;```

because I'd like to "promote" this proto to a dedicated directory - "metrics" - at the top-level of the gnostic repo. There I expect that we'll add other protos for other metrics that we'll add for API specs.

File: apps/vocabulary/vocabulary.pb.go [hunk 4]
```diff
@@ -0,0 +1,268 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
```
timburks, line 1: Please move this file and `vocabulary.proto` to a new top-level directory in the gnostic repo called `metrics`. 

To regenerate the file, please add this line to the end of `COMPILE-PROTOS.sh` at the top-level:

```protoc --go_out=. metrics/vocabulary.proto```

Then you can regenerate the file and import it from `apps/vocabulary/main.go`

File: apps/vocabulary/main.go [hunk 5]
```diff
@@ -0,0 +1,261 @@
...
+	//Start
```
noahdietz, line 52: Nit: please remove this comment or write a complete one. See https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences or https://golang.org/doc/effective_go.html#commentary for extra guidance.

apps/vocabulary/main.go
```diff
@@ -0,0 +1,261 @@
...
+	if document.Definitions != nil && document.Definitions.AdditionalProperties != nil {
```
noahdietz, line 53: The `document.Definitions.AdditionalProperties != nil` check is unnecessary. `range` treats a `nil` slice as an empty slice. For example: https://play.golang.org/p/EXetHoOv_Wq

Same goes for the other built-ins `len` and `cap` - they return `0` when given a `nil` slice.

apps/vocabulary/main.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+}
+
+func printSchema(schema *pb.Schema, properties map[string]int) {
+	if schema.Properties != nil {
+		for _, pair := range schema.Properties.AdditionalProperties {
+			addToMap(pair.Name, properties)
+		}
+	}
```
noahdietz, line 112: In Go, we prefer the "special case" logic to be indented and the main logic to be as far to the left as possible.

So in this case we'd invert the `if` condition check, early `return` in the condition body, otherwise fall through to the mainline logic of the `for` loop. 

Related guidance: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

apps/vocabulary/main.go
```diff
@@ -0,0 +1,261 @@
...
+
+func main() {
+	flag.Parse()
+	args := flag.Args()
+
+	document, err := readDocumentFromFileWithName(args[0])
+
+	if err != nil {
+		log.Printf("Error reading %s. This sample expects OpenAPI v2.", args[0])
+		os.Exit(-1)
```
noahdietz, line 123: Why exit `-1`? Is that a `gnostic` app thing? I would just use `1`.

apps/vocabulary/main.go
```diff
@@ -0,0 +1,261 @@
...
+			Count: int32(v),
+		}
+		vocab.Properties = append(vocab.Properties, temp)
+	}
+
+	bytes, err := proto.Marshal(vocab)
+	if err != nil {
+		panic(err)
+	}
+	err = ioutil.WriteFile("vocabulary_results.pb", bytes, 0644)
```
noahdietz, line 177: Suggestion: prefix the generated file with the spec file's name. Makes it easier to use this tool in scripts, because the user wont have to `mv` the generated results file to protect from being overwritten.

apps/vocabulary/main.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	document := &pb.Document{}
+	err = proto.Unmarshal(data, document)
+	if err != nil {
+		return nil, err
+	}
+	return document, nil
+}
+func addToMap(word string, mapName map[string]int) {
+	_, ok := mapName[word]
```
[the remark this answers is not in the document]
timburks: I'll chime in here: we want to keep them distinct so that we can detect and report on inconsistencies and actual usage.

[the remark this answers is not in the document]
noahdietz: SGTM!

apps/vocabulary/main.go
```diff
@@ -0,0 +1,261 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
timburks: Also we should use "LLC" not "Inc" for all new files - and might separately do a global update in all files.

apps/vocabulary/main.go
timburks: Thanks, I didn't know this! (this code was inherited from another example that I wrote)

gizzon:
> Hi @gizzon, nice to e-meet you. I might do reviews now and then. Great start here! Just a couple of comments.

Thank you!!

apps/vocabulary/main.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	return document, nil
+}
+func addToMap(word string, mapName map[string]int) {
+	_, ok := mapName[word]
+	if ok {
+		mapName[word] += 1
+	} else {
+		mapName[word] = 1
+	}
```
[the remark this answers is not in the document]
gizzon: That's a great trick, definitely going to use that from now on (:

apps/vocabulary/vocabulary.proto
```diff
@@ -0,0 +1,28 @@
...
+    string word = 1;
+    int32 count = 2;
+}
+
+message Vocabulary{
+    repeated WordCount schemas = 1;
+    repeated WordCount properties = 2;
+    repeated WordCount operations = 3;
+    repeated WordCount paramaters = 4;
+}
```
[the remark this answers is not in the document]
noahdietz: Apologies, I wasn't clear! Please add a new line to the _end of the file_, after the closing brace `}`

noahdietz:
LGTM, I'll let Tim Approve it

apps/vocabulary/main.go
```diff
@@ -0,0 +1,173 @@
...
+	return document, nil
+}
+func addToMap(word string, mapName map[string]int) {
+	mapName[word] += 1
+}
+
+func processDocument(document *pb.Document, schemas map[string]int, operationId map[string]int, names map[string]int, properties map[string]int) {
+	if document.Definitions != nil {
+		for _, pair := range document.Definitions.AdditionalProperties {
```
rodsimpson, line 46: consider moving contents of if to a separate block

apps/vocabulary/main.go
```diff
@@ -0,0 +1,173 @@
...
+			addToMap(pair.Name, schemas)
+			processSchema(pair.Value, properties)
+		}
+	}
+	for _, pair := range document.Paths.Path {
+		v := pair.Value
+		if v.Get != nil {
```
rodsimpson, line 53: consider using a switch statement

apps/vocabulary/main.go
```diff
@@ -0,0 +1,173 @@
...
+			processOperation(v.Delete, operationId, names)
+		}
+	}
+}
+
+func processOperation(operation *pb.Operation, operationId map[string]int, names map[string]int) {
+	if operation.OperationId != "" {
+		addToMap(operation.OperationId, operationId)
+	}
+	for _, item := range operation.Parameters {
```
rodsimpson, line 75: consider breaking out the nested blocks into separate methods

apps/vocabulary/main.go
```diff
@@ -0,0 +1,173 @@
...
+	if schema.Properties == nil {
+		return
+	}
+	for _, pair := range schema.Properties.AdditionalProperties {
+		addToMap(pair.Name, properties)
+	}
+}
+
+func main() {
+	flag.Parse()
```
rodsimpson, line 107: consider pulling some of this method out into separate methods - this one is getting a little too long

apps/vocabulary/main.go
timburks: These conditions aren't mutually exclusive - a path can have any/all of Get/Post/Put/Delete defined - so since switch statements don't fallthrough by default, I have a weak preference for this series of if statements.

apps/vocabulary/main.go
timburks: SGTM. Following existing patterns, this might moving lines 76-93 below into a "processParameter" function. 

Also, I think we should take Noah's suggestion and replace addToMap with direct updates to the map, e.g. `names[t2.BodyParameterName] += 1`

timburks:
Thank you!

PR #176, opened 2020-06-01
Vocabulary App for V2 and V3

gizzon:
This pull request builds upon my previous one by including functionality for v3 APIs. The structure now for the vocabulary app is as follows: main.go receives an API description, determines which API version it received and then calls one of the two following files: v2Processor.go or v3Processor.go. Both files follow the same flow but are specific to the structure of that API's version.

### DIFF
--- a/apps/vocabulary/main.go
+++ b/apps/vocabulary/main.go
@@ -23,8 +23,6 @@ import (
 
 	"github.com/golang/protobuf/jsonpb"
 	metrics "github.com/googleapis/gnostic/metrics"
-	openapi_v2 "github.com/googleapis/gnostic/openapiv2"
-	openapi_v3 "github.com/googleapis/gnostic/openapiv3"
 	"google.golang.org/protobuf/proto"
 )
 
@@ -38,12 +36,6 @@ func fillProtoStructures(m map[string]int) []*metrics.WordCount {
 		counts = append(counts, temp)
 	}
 	return counts
-}
-func runV2(document *openapi_v2.Document) {
-
-}
-func runV3(document *openapi_v3.Document) {
-
 }
 
 func main() {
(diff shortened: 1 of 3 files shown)


apps/vocabulary/main.go
```diff
@@ -144,7 +54,25 @@ func main() {
 	var properties map[string]int
 	properties = make(map[string]int)
 
-	processDocument(document, schemas, operationId, names, properties)
+	//Temporary, for now using filename to check file type
+	version_flag := strings.Contains(args[0], "swagger")
```
timburks, line 58: LGTM. To other reviewers: we're planning to migrate this code into a plugin, and when we do, the plugins framework will tell us what version of spec we're getting so we won't need to switch on anything arbitrary.

apps/vocabulary/v2Processor.go
```diff
@@ -0,0 +1,91 @@
...
+	document := &openapi_v2.Document{}
+	err = proto.Unmarshal(data, document)
+	if err != nil {
+		return nil, err
+	}
+	return document, nil
+
+}
+
+func processDocumentV2(document *openapi_v2.Document, schemas map[string]int, operationId map[string]int, names map[string]int, properties map[string]int) {
```
rodsimpson, line 25: Not sure what the standard is for Go, so I'm just wondering if these arguments should break down to 1 arg per line(since the line is so long).

rodsimpson: here and for the other methods with many arguments

noahdietz, line 25: When there are sequential function arguments that share a type, they can be defined together like so:

func processDocumentV2(document *openapi_v2.Document, schemas, operationId, names, properties map[string]int)

Please do that here and elsewhere.

noahdietz: Running gofmt -w . in the directory will “do the right thing”.

gizzon: Thanks for suggesting this! I'm going to address it in the next PR

PR #179, opened 2020-06-02
gnostic-complexity: new plugin that computes simple complexity metrics

timburks:
This PR adds a new plugin that computes simple complexity metrics for an API that include the number of methods (of each type), schemas, and schema parameters. Metrics are described by a new "standard" proto in gnostic/metrics that can be used to parse complexity metrics in external related tools.


timburks:
@noahdietz @LorenzHW I wanted you to see the new metrics directory at the top level - its purpose is to contain protos for various measurements that we might make related to API specs. This PR adds a simple "complexity" metric, and separately @gizzon added a "vocabulary" metric that lists the words used in an API spec. Metrics are computed with plugins, and the protos are here (and public) so that external code can be written to access/display these metrics.

timburks:
Merging. @LorenzHW @noahdietz PTAL at your convenience. Thanks!

PR #181, opened 2020-06-04
Summarize vocabulary

gizzon:
The summarize vocabulary app runs the vocabulary plugin on multiple pb files via stdin. This app combines the protocol buffer for each individual spec into one large spec. The program will write to either one or both of the provided outputs: .csv and .pb.


apps/summarize-vocabulary/main.go
```diff
@@ -0,0 +1,156 @@
...
+
+	metrics "github.com/googleapis/gnostic/metrics"
+)
+
+var schemas map[string]int
+var operationId map[string]int
+var parameters map[string]int
+var properties map[string]int
+
+func pbToCSV(vocab *metrics.Vocabulary) {
```
timburks, line 35: let's name this "writeCSV"

apps/summarize-vocabulary/main.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	for _, s := range vocab.Parameters {
+		temp := fmt.Sprintf("%s,%s,%d\n", "parameters", s.Word, int(s.Count))
+		f4.WriteString(temp)
+	}
+	f4.Close()
+}
+
+func pbOutput(combinedVocab *metrics.Vocabulary) {
+	bytes, err := proto.Marshal(combinedVocab)
```
timburks, line 62: to parallel the function above, I'd like to call this "writePb"

apps/summarize-vocabulary/main.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		panic(err)
+	}
+	err = ioutil.WriteFile("summarize-vocabulary.pb", bytes, 0644)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func openCombinedPbResults() *metrics.Vocabulary {
```
timburks, line 72: I think this function could also be named more descriptively. How about "combineVocabularies"?

apps/summarize-vocabulary/main.go
```diff
@@ -0,0 +1,156 @@
...
+		properties[prop.Word] += int(prop.Count)
+	}
+}
+
+func main() {
+	csvPtr := flag.Bool("csv", false, "generate csv output")
+	pbPtr := flag.Bool("pb", false, "generate pb output")
+
+	flag.Parse()
+
```
timburks, line 142: If neither --csv or --pb is provided, we should print a help message and exit. I think this function will do that: https://golang.org/pkg/flag/#PrintDefaults

apps/summarize-vocabulary/main.go
```diff
@@ -0,0 +1,156 @@
...
+		f4.WriteString(temp)
+	}
+	for _, s := range vocab.Parameters {
+		temp := fmt.Sprintf("%s,%s,%d\n", "parameters", s.Word, int(s.Count))
+		f4.WriteString(temp)
+	}
+	f4.Close()
+}
+
+func pbOutput(combinedVocab *metrics.Vocabulary) {
```
timburks, line 61: "combinedVocab" could just be "v" (see more thoughts about variable names below)

apps/summarize-vocabulary/main.go
```diff
@@ -0,0 +1,156 @@
...
+
+func readVocabularyFromFileWithName(filename string) {
+	data, err := ioutil.ReadFile(filename)
+	if err != nil {
+		fmt.Printf("File error: %v\n", err)
+		os.Exit(1)
+	}
+
+	vocab := &metrics.Vocabulary{}
+	err = proto.Unmarshal(data, vocab)
```
timburks, line 115: I think "v" is good to use instead of "vocab". Here is some related discussion https://www.quora.com/Why-does-Golang-promote-short-and-kind-of-meaningless-names-for-variables

Looking through the file, I think that's generally true for all the instances of "vocab". None of these variables live more than a few lines, so reading Ian Lance Taylor's advice (in that quora link), I think I'd prefer that they all be single-letter variables.

apps/summarize-vocabulary/main.go
```diff
@@ -0,0 +1,156 @@
...
+	"io/ioutil"
+	"os"
+	"sort"
+
+	"github.com/golang/protobuf/proto"
+
+	metrics "github.com/googleapis/gnostic/metrics"
+)
+
+var schemas map[string]int
```
timburks, line 30: please put a comment above line 30 that explains why these variables are global

PR #190, opened 2020-06-09
vocabulary-operations

gizzon:
This new app provides the ability to run operations of Vocabulary structures. The implemented operations are: difference, intersection, union and export.


apps/vocabulary-operations/intersection.go
```diff
@@ -0,0 +1,60 @@
+package main
+
+import (
+	metrics "github.com/googleapis/gnostic/metrics"
+)
+
+func mapIntersection(v *metrics.Vocabulary) {
```
rodsimpson, line 7: This method looks identical to one in main.go - should we try to consolidate these?

apps/vocabulary-operations/main.go
```diff
@@ -0,0 +1,92 @@
...
+	if !*unionPtr && !*intersectionPtr && !*differencePtr && !*exportPtr {
+		flag.PrintDefaults()
+		fmt.Printf("Please use one of the above command line arguments.\n")
+		os.Exit(-1)
+		return
+	}
+	v := make([]*metrics.Vocabulary, 0)
+	switch arguments := len(args); arguments {
+	case 0:
+		files := openVocabularyFiles()
```
rodsimpson, line 66: consider moving the functionality inside the case and default blocks to their own methods. could make this large-ish method a little easier to read.

apps/vocabulary-operations/intersection.go
gizzon: I first created intersection in the vocabulary-difference app. Tim and I had the idea to build upon it and create other operators which are now all in the vocabulary-operations app. I believe in my next pull I'll be deleted vocabulary-difference so there shouldn't be any more duplicates

apps/vocabulary-operations/main.go
```diff
@@ -86,10 +86,11 @@ func main() {
 	intersectionPtr := flag.Bool("intersection", false, "generates the intersection of pb files")
 	differencePtr := flag.Bool("difference", false, "generates the difference of pb files")
 	exportPtr := flag.Bool("export", false, "export a given pb file as a csv file")
+	companyPtr := flag.Bool("company", false, "egenerates uniqueness within company")
```
timburks, line 89: Please update this to "filterCommon" and make the description match.
I think the flag should be "--filter-common" to match Unix command line conventions.

metrics/vocabulary/vocabulary_test.go
```diff
@@ -57,186 +47,98 @@ func testVocabulary(t *testing.T, operation func(a []*metrics.Vocabulary) *metri
...
-		Properties: fillTestProtoStructue(propWords2, propCount2),
-		Operations: fillTestProtoStructue(opsWords2, opsCount2),
-		Parameters: fillTestProtoStructue(paramWords2, paramCount2),
+		Schemas:    fillTestProtoStructue([]string{"Hello", "random", "status", "google"}, []int{5, 6, 1, 4}),
+		Properties: fillTestProtoStructue([]string{"cat", "dog", "thing"}, []int{4, 3, 2}),
+		Operations: fillTestProtoStructue([]string{"countPrint", "print", "funcName"}, []int{17, 12, 19}),
+		Parameters: fillTestProtoStructue([]string{"name", "id", "tag", "suggester"}, []int{5, 1, 1, 15}),
 	}
 
 	vocab := make([]*metrics.Vocabulary, 0)
```
timburks, line 64: I think this slice (and similar ones) should be named "vocabs" or "vocabularies" to clearly show that they are collections and are not individual vocabularies.

PR #191, opened 2020-06-18
Linter App

gizzon:
This PR includes a new app, parseLinter. ParseLinter accepts output from either the IBM Linter or Spectral Linter. The output is then packaged into a Linter pb and written to a file. This PR also includes improvements to the vocabulary testing that supports the new structure. Coverage is now at 81%


metrics/vocabulary/vocabulary.go
```diff
@@ -70,12 +70,14 @@ func WriteCSV(v *metrics.Vocabulary, filename string) {
 func WritePb(v *metrics.Vocabulary) {
 	bytes, err := proto.Marshal(v)
```
timburks, line 71: Please make this function return err (instead of printing errors and exiting)

timburks: thanks!

PR #192, opened 2020-06-30
Created a mvp API Linter plugin that supports both v2 and v3

gizzon:
This pull request includes a new feature for the gnostic repo, an API Linter. This linter supports both v2 and v3 APIs. Currently, the linter has only a few rules to check for but will be added to in the future. The linter plugin accepts one API and will return a protobuf which contains the output. The structure for the output is as follows:
Type ("Error" or "Warning")
Message
Suggestion (if applicable)
Path

I am currently investigating how to include line numbers which will hopefully be included soon.


metrics/lint/aip-linterv2.go
```diff
@@ -0,0 +1,104 @@
...
+				case *pb.NonBodyParameter_QueryParameterSubSchema:
+					parameters = append(parameters, rules.Field{Name: t3.QueryParameterSubSchema.Name, Path: path})
+				}
+			}
+		}
+	}
+	return parameters
+}
+
+func AIPLinterV2(document *pb.Document) (*Linter, int) {
```
timburks, line 91: I'd like to generally use verbs for function names and nouns (like "AIPLinterV2") for structs. For more discussion of Go naming, see [this blog post](http://codeblooded.me/2015/08/20/lessons-in-golang-naming.html).

Based on that, I think this function would be better named "AIPLintV2".

metrics/lint/aip-linterv3.go
```diff
@@ -0,0 +1,110 @@
...
+			}
+			if v.Delete != nil {
+				p = append(p, processOperationV3(v.Delete, path)...)
+			}
+		}
+	}
+	return p
+}
+
+func AIPLinterV3(document *openapi_v3.Document) (*Linter, int) {
```
timburks, line 97: See naming comments above.

metrics/lint/linter.proto
```diff
@@ -0,0 +1,32 @@
...
+
+
+syntax = "proto3";
+package linter;
+
+// The Go package name.
+option go_package = "linter";
+
+message Message{
+    string Type = 1;
```
timburks, line 23: From the [Protocol Buffers Style Guide](https://developers.google.com/protocol-buffers/docs/style#message_and_field_names), field names should be lower-cased.

metrics/lint/linter.proto
```diff
@@ -0,0 +1,32 @@
...
+    string Message = 2;
+    string Suggestion = 3;
+    repeated string Path = 4;
```
timburks, line 26: From the [Protocol Buffers Style Guide](https://developers.google.com/protocol-buffers/docs/style#repeated_fields), repeated fields should be plural. Since "paths" doesn't really make sense (we're describing a single path), please use "keys" for this field.

metrics/lint/linter.proto
```diff
@@ -0,0 +1,32 @@
...
+    int32 Line = 5;
+}
+
+message Linter{
+    repeated Message linterResults = 1;
```
timburks, line 31: I think this should be "messages" instead of "linterResults" - this seems more predictable and it avoids repetition of the word "linter".

metrics/lint/openapi-validator.go
```diff
@@ -0,0 +1,168 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"google.golang.org/protobuf/proto"
+)
+
+type Lint struct {
```
timburks, line 13: Please add a comment here (above this struct) that explains where these definitions come from and link to their source - this might be code or documentation in the IBM linter project.

timburks: Also, if this is specific to the IBM linter, could it be namespaced appropriately? 

Both this and the spectral linter support might be better in their own subpackages of metrics/lint

plugins/gnostic-linter/main.go
```diff
@@ -0,0 +1,102 @@
...
+	env, err := plugins.NewEnvironment()
+	env.RespondAndExitIfError(err)
+
+	var linter *lint.Linter
+	f, err := os.OpenFile("results.txt",
+		os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		log.Println(err)
+	}
+	defer f.Close()
```
timburks, line 55: Plugins should return all results back through the [plugin interface](https://github.com/googleapis/gnostic/blob/ae26236156ff98d8de9d2bad34d4103b487b0a9a/plugins/plugin.proto#L135). 

It looks like this is dead code... if so, please remove it.

plugins/gnostic-linter/main.go
```diff
@@ -0,0 +1,102 @@
...
+			documentv2 := &openapiv2.Document{}
+			err = proto.Unmarshal(model.Value, documentv2)
+			if err == nil {
+				// Analyze the API document.
+				linter, _ = lint.AIPLinterV2(documentv2)
+				// num := strconv.Itoa(numErrors)
+				// if _, err := f.WriteString(env.Request.SourceName + "," + num + "\n"); err != nil {
+				// 	log.Println(err)
+				// }
+			}
```
timburks, line 69: please remove this commented-out code

plugins/gnostic-linter/main.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if linter != nil {
+		outputName1 := filepath.Join(
+			filepath.Dir(env.Request.SourceName), "linter.json")
+		outputName2 := filepath.Join(
+			filepath.Dir(env.Request.SourceName), "linter.pb")
+		file := &plugins.File{}
+
+		file.Name = outputName1
+		file.Data, err = json.MarshalIndent(linter, "", "  ")
```
timburks, line 88: please remove the outputName1 (and outputName2) variables and put their values inline here.

timburks: Lines 85 and following should as much as possible look like lines 93 and following - same formatting, line spacing, etc.

metrics/rules/aip140_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+package rules
+
+import (
+	"testing"
+)
+
+func TestSnakeCase(t *testing.T) {
+	if pass, suggestion := snakeCase("helloWorld"); pass {
+		t.Error("Given \"helloWorld\", snakeCase() returned true, expected false")
```
timburks, line 23: These tests seem to replicate a lot of code. I think it would be helpful to restructure them to start by defining an array of test strings and then looping over those strings calling the code in lines 22-26 (for this function, similarly for the others). That would make it easier to add new patterns to each test or to change common code used by all the tests.

metrics/rules/aip140.go
```diff
@@ -0,0 +1,133 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// AIP140Driver calls all functions for AIP rule 140
+func AIP140Driver(f Field) []MessageType {
+	messages := make([]MessageType, 0)
+	val, sugg := snakeCase(f.Name)
```
timburks, line 92: These function names don't say much about what is happening. Referring back to the suggestion that function names should always begin with verbs, perhaps these should be called "checkSnakeCase", "checkAbbreviations", etc.

metrics/rules/aip140.go
```diff
@@ -0,0 +1,133 @@
...
+	if !val {
+		m := []string{"Error", "Parameter names must follow case convention: lower_snake_case\n",
+			fmt.Sprintf("Rename field %s to %s\n", f.Name, sugg)}
+		temp := MessageType{Message: m, Path: f.Path}
+		messages = append(messages, temp)
+	}
+	val, sugg = abbreviation(f.Name)
```
timburks, line 99: Just noticing that "abbreviation" is singular while "numbers", "reservedWords", and "prepositions" are plural... should it also be plural?

gizzon: I don't believe so in this case. This function checks if the parameter is an abbreviation, not if there exists multiple abbreviations within the name.

Issue #31, opened 2017-03-01
Report errors by line number in source documents.


timburks:
Currently errors are reported in terms of key paths in the source documents. It would be helpful to users to have this reflected back to line numbers.

MikeRalphson:
Perhaps there is a go equivalent of [this](https://github.com/epoberezkin/json-source-map) library, or it could be ported?

timburks:
It appears that the [yaml.v3](https://ubuntu.com/blog/api-v3-of-the-yaml-package-for-go-is-available) [Node](https://godoc.org/gopkg.in/yaml.v3#Node) will give us line numbers. Thanks @gizzon for the find!

timburks:
Fixed in #219

PR #194, opened 2020-07-12
Update gnostic to use yaml.v3

timburks:
This set of changes upgrades gnostic's JSON/YAML reading to use gopkg.in/yaml.v3.

In addition to having a cleaner interface, yaml.v3 maintains line and column numbers for all parsed elements, which will allow us to address #31 in a future set of changes.

This first set of commits should be enough for all tests to pass, but there are still many rough edges to be addressed before this is merged.


openapiv3/OpenAPIv3.proto
```diff
@@ -208,8 +208,8 @@ message Info {
   Contact contact = 4;
   License license = 5;
   string version = 6;
-  repeated NamedAny specification_extension = 7;
-  string summary = 8;
+  string summary = 7;
+  repeated NamedAny specification_extension = 8;
```
timburks, line 212: These need to be swapped back

testdata/library-example-with-ext.text.out
```diff
@@ -195,8 +195,8 @@ vendor_extension: <
   name: "x-sampleone-mysimpleint64"
   value: <
     value: <
-      type_url: "type.googleapis.com/google.protobuf.StringValue"
-      value: "\n\00512345"
+      type_url: "type.googleapis.com/google.protobuf.Int64Value"
+      value: "\010\271`"
     >
```
timburks, line 200: This test data was updated because the original was incorrect.

extensions/sample/x-sampleone.json
```diff
@@ -47,7 +47,7 @@
                     "id": "x-sampleone-mysimpleboolean"
                 },
                 "PrimitiveInt64": {
-                    "type": "string",
```
timburks, line 50: This test was incorrect.

openapiv3/OpenAPIv3.proto
timburks: Fixed.

timburks:
This is ugly in places but all tests pass and #195 provides a fix for clients that I expect will be broken by this change.

PR #198, opened 2020-07-16
Adding an additional case for ScalarNodes with tag "!!int".

LorenzHW:
Tests in gnostic-grpc are currently failing due to [PR194](https://github.com/googleapis/gnostic/pull/194).
Currently, `ResponseOrRefernce` is not properly constructed, because gnostic does [not recognize](https://github.com/googleapis/gnostic/blob/ce2874882b96283118ceb88bf3b651e37636963e/openapiv3/OpenAPIv3.go#L3726) that a yaml node can also have the tag "!!int".

This PR fixes all test cases in gnostic-grpc.


timburks:
Thanks!

PR #200, opened 2020-07-17
Use key paths to get source info from yaml files

gizzon:
…or a field in yaml files


metrics/line-numbers/fineLineNumber.go
```diff
@@ -0,0 +1,48 @@
+package linenumbers
```
timburks, line 1: I think this file's name has a typo.

... regardless of that, I'd like to use a different name for this package. yaml.Nodes give us access to comments and column positions along with line numbers, and this recursive is useful for getting all of that.

Protocol Buffers call this kind of information "source info" and I think that's a good pattern to follow. Let's name this package "sourceinfo" (and put it in a directory named "metrics/sourceinfo"). This file could be just "sourceinfo.go".

metrics/line-numbers/fineLineNumber.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strconv"
+
+	"gopkg.in/yaml.v3"
+)
+
+func recursiveFindLine(node *yaml.Node, keyIndex int, maxDepth int, keys []string) (*yaml.Node, error) {
```
timburks, line 11: I don't think it's necessary to include "recursive" in the names of recursive functions. Since this function is returning a node and not a line number, I think "findNode" would be a good name to use for this.

metrics/line-numbers/fineLineNumber.go
```diff
@@ -0,0 +1,48 @@
...
+		} else {
+			continue
+		}
+
+	}
+	return &yaml.Node{}, nil
+}
+
+// FindYamlLine returns a line number for a field within a yaml file given the path and yaml file.
+func FindYamlLine(filename string, keys []string, token string) (*yaml.Node, error) {
```
timburks, line 38: Let's call this "FindNode" - this public interface would not conflict with the private name of the function (above) and the name similarity would help make their relationship clear.

metrics/line-numbers/lineNumbers_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"testing"
+)
+
+//TestFindLineNumbers runs unit tests on the linenumbers package
+func TestFindLineNumbers(t *testing.T) {
+	keys := []string{"paths", "/alerts?lat={lat}\u0026lon={lon}", "get", "parameters", "1", "name"}
+	token := "lon_100"
+	file := "../../examples/v2.0/yaml/weatherbit.yaml"
```
timburks, line 11: I'd like to avoid including third-party (or at least non-Google) API descriptions in the gnostic repo - ownership and usage rights are unclear. Could you modify this test to instead use the petstore.yaml specs that are already in the repo (in the examples directory)? If so, please also remove weatherbit.yaml from this PR.

metrics/sourceinfo/sourceinfo_test.go
```diff
@@ -0,0 +1,54 @@
+package sourceinfo
+
+import (
+	"testing"
+)
+
+//TestFindLineNumbers runs unit tests on the linenumbers package
```
timburks, line 7: Let's update this comment to the new package name (or remove the phrase "on the linenumbers package")

PR #203, opened 2020-07-20
Refined vocabulary to use structs instead of global variables

gizzon:


metrics/vocabulary/openapiv3.go
```diff
@@ -25,47 +39,47 @@ func fillProtoStructures(m map[string]int) []*metrics.WordCount {
 	return counts
 }
 
-func processOperationV3(operation *openapi_v3.Operation, operationID, parameters map[string]int) {
+func processOperationV3(operation *openapi_v3.Operation, vocab *Vocabulary) {
```
timburks, line 42: Whenever you have a function that takes vocab *Vocabulary as an argument, please make this a function on the Vocabulary struct. That would change the function declaration to look like this:

func (vocab *Vocabulary) processOperationV3(operation *openapi_v3.Operation) {

timburks: You would typically call this with:
```
vocab.processOperation..
```

PR #206, opened 2020-07-31
crude and simple tool for building vocabularies from proto files.

timburks:
* for review only *

This builds on the vocabulary work by creating vocabulary structures for .proto files. Several things were copied-and-pasted from metrics/vocabulary because they had private scope and couldn't be used outside the package. We should either make those fields public or put the proto vocabulary generation in the vocabulary package (my concern about that is that it adds a dependency on go-protoparser).

To try it, build `proto-vocabulary`  and run it on a directory of protos with
```
% proto-vocabulary --path <path to directory containing protos>
```
or on a single proto with 
```
% proto-vocabulary --proto <path to proto>
```

It will produce a single csv file, `vocabulary.csv`.

### DIFF
--- a/apps/proto-vocabulary/main.go
+++ b/apps/proto-vocabulary/main.go
@@ -1,0 +1,174 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	metrics "github.com/googleapis/gnostic/metrics"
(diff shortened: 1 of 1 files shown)


timburks:
This has been merged into another project, to-be-released soon. I'll close this when that becomes available.

timburks:
Vocabulary analysis for protos is now available in https://github.com/apigee/registry.

PR #216, opened 2020-10-02
Fix checking of error from jsonwriter.Marshal

bvwells:
Fix checking of error from jsonwriter.Marshal.


timburks (approved):
Thanks for fixing this!

bvwells:
Thanks @timburks for the review and merge!

PR #221, opened 2020-10-23
protoc-gen-openapi

timburks:
This PR adds a protoc plugin that generates an OpenAPI v3 YAML description of a transcoded gRPC API following the conventions described in [AIP-127](https://google.aip.dev/127). So far it's been verified on a couple of examples (one public, one private) so it's likely to need improvements to handle all possible cases. One known limitation: The API name and description are taken from the "service" section, so if a set of compiled protos includes multiple service sections, an arbitrary name and description will be selected.

This generates OpenAPI v3 only. There are no plans to produce OpenAPI v2 output.

For example usage, see the test in apps/protoc-gen-openapi/plugin_test.go.


LorenzHW:
That's pretty cool Tim! Now it is possible to do an endless loop with gnostic-grpc 🔄

noahdietz (approved):

timburks:
@noahdietz Thanks for reviewing and for pointing out the v2 plugin API!

@LorenzHW Yes, we should try round-tripping!

PR #223, opened 2020-11-13
protoc-gen-openapi: timestamps should be serialized as strings

timburks:


LEI:
Could this be applied to `Date` and `DateTime` and their [corresponding format](https://swagger.io/docs/specification/data-models/data-types/#format) as well?

I expected the timestamp to be respresented as an integer but it is only my opinion.

PR #234, opened 2020-12-06
protoc-gen-openapi: serialize date and date-time as strings

LEI:
The OpenAPI Specification defines the `date` and `date-time` [DataTypes](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#data-types) based on the [JSON Schema built-in formats](https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times). For reference the following issue centralizes discussions about this topic: OAI/OpenAPI-Specification#845.

This adds [`google/type/date.proto`](https://github.com/googleapis/googleapis/blob/master/google/type/date.proto) and [`google/type/datetime.proto`](https://github.com/googleapis/googleapis/blob/master/google/type/datetime.proto) under `apps/protoc-gen-openapi/examples`.


timburks (approved):
Thank you!

PR #244, opened 2021-04-27
Update protoc-gen-openapi to represent map<string,string> as OpenAPI objects.

timburks:
As illustrated in the new test, this change causes map fields of the form:
```
map<string, string> labels = 2;
```
to be represented as "objects" in OpenAPI.
```
labels:
    type: object
```

This change was made to correctly generate https://github.com/apigee/registry/blob/main/openapi.yaml

### DIFF
--- a/apps/protoc-gen-openapi/plugin_test.go
+++ b/apps/protoc-gen-openapi/plugin_test.go
@@ -30,14 +30,12 @@ func TestLibraryOpenAPI(t *testing.T) {
 		"examples/google/example/library/v1/library.proto",
 		"--openapi_out=.").Run()
 	if err != nil {
-		t.Logf("protoc failed: %+v", err)
-		t.FailNow()
+		t.Fatalf("protoc failed: %+v", err)
 	}
 	// Verify that the generated spec matches our expected version.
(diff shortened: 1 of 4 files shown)


timburks (commented):
@seaneganx could you take a look at this?

apps/protoc-gen-openapi/plugin_test.go
```diff
@@ -65,3 +65,26 @@ func TestBodyMappingOpenAPI(t *testing.T) {
...
+	var err error
+	// Run protoc and the protoc-gen-openapi plugin to generate an OpenAPI spec.
+	err = exec.Command("protoc",
+		"-I", "../../",
+		"-I", "../../third_party",
+		"-I", "examples",
+		"examples/tests/mapfields/message.proto",
+		"--openapi_out=.").Run()
+	if err != nil {
+		t.Logf("protoc failed: %+v", err)
```
seaneganx, line 79: Nit: `t.Fatalf` is equivalent to `t.Logf` followed by `t.FailNow`. Both occurrences can be updated.

seaneganx (approved):

apps/protoc-gen-openapi/plugin_test.go
timburks: Thanks! I've updated all six instances in that test.

PR #259, opened 2021-09-26
Fix module name and support goimports local prefix

zchee:
Fix module name to `github.com/google/gnostic`, and support goimports local prefix.


timburks (approved):
@zchee Looks great! Thank you very much!

PR #266, opened 2021-11-03
Suggested fixes

micmorris:
- Output schemas with no other type as 'object'
- Output field names as camelCase instead of snake_case
- Make operationId just the method name


micmorris:
Oh whoops, this was supposed to be against my fork, it's not ready yet.

PR #271, opened 2021-11-26
protoc-gen-openapi: Convert maps and lists of sub messages

morphar:
This ensures proper conversion of maps and lists of standard types, (some) google.protobuf types and sub messages.

### DIFF
--- a/apps/protoc-gen-openapi/examples/tests/mapfields/openapi.yaml
+++ b/apps/protoc-gen-openapi/examples/tests/mapfields/openapi.yaml
@@ -36,6 +36,8 @@ components:
             properties:
                 message_id:
                     type: string
+                sub_message:
+                    $ref: '#/components/schemas/SubMessage'
                 string_list:
                     type: array
                     items:
@@ -49,11 +51,17 @@ components:
                     items:
                         type: object
                 strings_map:
-                    type: string
+                    type: object
+                    additionalProperties:
+                        type: string
                 sub_messages_map:
-                    $ref: '#/components/schemas/SubMessage'
+                    type: object
+                    additionalProperties:
+                        $ref: '#/components/schemas/SubMessage'
                 objects_map:
                     type: object
+                    additionalProperties:
+                        type: object
         SubMessage:
             properties:
                 id:
--- a/apps/protoc-gen-openapi/examples/tests/mapfields/openapi.yaml
+++ b/apps/protoc-gen-openapi/examples/tests/mapfields/openapi.yaml
@@ -36,7 +36,30 @@ components:
             properties:
                 message_id:
                     type: string
-                labels:
+                string_list:
+                    type: array
+                    items:
+                        type: string
+                sub_message_list:
+                    type: array
+                    items:
+                        $ref: '#/components/schemas/SubMessage'
+                object_list:
+                    type: array
+                    items:
+                        type: object
+                strings_map:
+                    type: string
+                sub_messages_map:
+                    $ref: '#/components/schemas/SubMessage'
+                objects_map:
                     type: object
+        SubMessage:
+            properties:
+                id:
+                    type: integer
+                    format: int64
+                label:
+                    type: string
 tags:
     - name: Messaging
--- a/apps/protoc-gen-openapi/examples/tests/mapfields/message.proto
+++ b/apps/protoc-gen-openapi/examples/tests/mapfields/message.proto
@@ -38,10 +38,11 @@ message SubMessage {
 
 message Message {
     string message_id = 1;
-    repeated string string_list = 2;
-    repeated SubMessage sub_message_list = 3;
-    repeated google.protobuf.Struct object_list = 4;
-    map<string, string> strings_map = 5;
-    map<string, SubMessage> sub_messages_map = 6;
-    map<string, google.protobuf.Struct> objects_map = 7;
+    SubMessage sub_message = 2;
+    repeated string string_list = 3;
+    repeated SubMessage sub_message_list = 4;
+    repeated google.protobuf.Struct object_list = 5;
+    map<string, string> strings_map = 6;
+    map<string, SubMessage> sub_messages_map = 7;
+    map<string, google.protobuf.Struct> objects_map = 8;
 }
--- a/apps/protoc-gen-openapi/examples/tests/mapfields/message.proto
+++ b/apps/protoc-gen-openapi/examples/tests/mapfields/message.proto
@@ -18,6 +18,7 @@ syntax = "proto3";
 package tests.mapfields.message.v1;
 
 import "google/api/annotations.proto";
+import "google/protobuf/struct.proto";
 
 option go_package = "github.com/google/gnostic/apps/protoc-gen-openapi/examples/tests/mapfields/message/v1;message";
 
@@ -29,7 +30,18 @@ service Messaging {
         };
     }
 }
+
+message SubMessage {
+    int64 id = 1;
+    string label = 2;
+}
+
 message Message {
     string message_id = 1;
-    map<string, string> labels = 2;
+    repeated string string_list = 2;
+    repeated SubMessage sub_message_list = 3;
+    repeated google.protobuf.Struct object_list = 4;
+    map<string, string> strings_map = 5;
+    map<string, SubMessage> sub_messages_map = 6;
+    map<string, google.protobuf.Struct> objects_map = 7;
 }
--- a/apps/protoc-gen-openapi/examples/tests/mapfields/openapi_json.yaml
+++ b/apps/protoc-gen-openapi/examples/tests/mapfields/openapi_json.yaml
@@ -36,7 +36,30 @@ components:
             properties:
                 messageId:
                     type: string
-                labels:
+                stringList:
+                    type: array
+                    items:
+                        type: string
+                subMessageList:
+                    type: array
+                    items:
+                        $ref: '#/components/schemas/SubMessage'
+                objectList:
+                    type: array
+                    items:
+                        type: object
+                stringsMap:
+                    type: string
+                subMessagesMap:
+                    $ref: '#/components/schemas/SubMessage'
+                objectsMap:
                     type: object
+        SubMessage:
+            properties:
+                id:
+                    type: integer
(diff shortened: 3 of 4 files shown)


morphar:
Hi again @timburks :)
I had actually planned to get this change into the previous pull request, but actually forgot about it.
There's a bit of code cleanup, but most importantly: proper handling of sub messages.

apps/protoc-gen-openapi/examples/tests/mapfields/openapi.yaml
```diff
@@ -36,7 +36,30 @@ components:
...
+                        type: string
+                sub_message_list:
+                    type: array
+                    items:
+                        $ref: '#/components/schemas/SubMessage'
+                object_list:
+                    type: array
+                    items:
+                        type: object
+                strings_map:
```
timburks, line 51: Should this (and the two maps below) be an object?

timburks (commented):
@morphar Thanks for continuing to drive this forward! See my question below about expected handling of maps - if it's unclear we could set up a sample API and an Envoy proxy to see how its gRPC transcoding handles maps.  

Update: I have an API that has [message fields of type `map<string,string>`](https://github.com/apigee/registry/blob/d783b46162f0ba82bd3bf948232892bddb654669/google/cloud/apigeeregistry/v1/registry_models.proto#L80). When I call it using Envoy and transcoding, I'm able to set those fields with JSON objects like this:

`"labels": {"foo": "123", "bar": "456"}`

When I change a field type from string to integer, I get errors. For this:

`"labels": {"foo": 123, "bar": "456"}`

I get this transcoding error: 

`api.labels[0].value: invalid value 123 for type TYPE_STRING`

I don't have an API with maps of messages (if needed, we could construct a sample) but I suspect/hope that similar type checking would occer for those.

apps/protoc-gen-openapi/examples/tests/mapfields/openapi.yaml
morphar: Good catch!
Sorry... I went the wrong way around while coding this.
My thinking was "how to express map protos in OpenAPI?" - instead of: is this even doable in OpenAPI and how would it look.

I tried to get the code to parse sub messages and then I changed the example yaml files, once it was working...
Clearly that was not a good idea.
 
I think object_list and the other lists is correct.
`repeated google.protobuf.Struct object_list = 4;` should be:
```yaml
object_list:
    type: array
    items:
        type: object
```
It's an array (repeated) of objects (Struct) with no restrictions, right?

I think that the correct version of the maps is something like this:
```yaml
  stringsMap:
      type: object
      additionalProperties:
        type: string
  subMessagesMap:
      type: object
      additionalProperties:
        $ref: '#/components/schemas/SubMessage'
  objectsMap:
      type: object
      additionalProperties:
        type: object
```
The `object` type, naturally forces key to be a string.
The additionalProperties restricts the value of each property.
This seems to produce the correct example values in Swagger editor:
```json
{
  "messageId": "string",
  "stringList": [
    "string"
  ],
  "subMessageList": [
    {
      "id": 0,
      "label": "string"
    }
  ],
  "objectList": [
    {}
  ],
  "stringsMap": {
    "additionalProp1": "string",
    "additionalProp2": "string",
    "additionalProp3": "string"
  },
  "subMessagesMap": {
    "additionalProp1": {
      "id": 0,
      "label": "string"
    },
    "additionalProp2": {
      "id": 0,
      "label": "string"
    },
    "additionalProp3": {
      "id": 0,
      "label": "string"
    }
  },
  "objectsMap": {
    "additionalProp1": {},
    "additionalProp2": {},
    "additionalProp3": {}
  }
}
```

Maybe I should try to get map<int32,string> and the likes to be converted to arrays - if Envoy can handle that.

I will make the changes and push again.

morphar:
@timburks You're very welcome! I'm just contributing to an awesome project, that saves me a lot of time ;)

I'm working on an API with Envoy in front, a gRPC backend in Go, a frontend in Vue.js 3, which uses the protoc generated js and grpc-web code and a Postman setup for checking the REST API part.
I use [https://editor.swagger.io](https://editor.swagger.io) for testing the OpenAPI results and checking ideas for modifications.
So I can check how everything interacts and see what happens with regards to map<string, SubMessage> and the likes.

morphar:
I have fixed the maps.
I haven't tested it through the entire eco-system yet, but that is next on the list.
If I find any issues with Envoy or any other part, I will make a fix.
But as far as I can tell, this should all be straight forward now.
The OpenAPI code makes Swagger Editor produce correct example objects and the rest should be "simple" json transcoding.

apps/protoc-gen-openapi/examples/tests/mapfields/openapi.yaml
timburks: Thanks, I agree that using additionalProperties is a good way to constrain the type of map elements (I'd forgotten about that).

timburks:
Looks great! If I can give you [one more PR](https://github.com/morphar/gnostic/pull/1), I have a fix for a problem that I found running this on the Google Translate API (v3) - that API has messages nested inside other messages, and the nested messages weren't being represented with OpenAPI schemas. That change partially fixes that - it will produce incorrect results when messages with the same name are defined inside different messages (naming collisions), but it might be enough for Google APIs. [Here's a gist](https://gist.github.com/timburks/44f3671a7e80b0cd7a9b922192f3941b) containing a version of the generator that generates hierarchical names, but it's messy and I (temporarily) dropped it when I saw that the Discovery doc contained non-hierarchical names. I think we'll want to come back to this, but I think we should hold the change in the gist for a future PR.

morphar:
Thanks Tim! I have merged your pull request 👍 
I will have a look at the Gist now :)

morphar:
Just a thought with regards to the nested messages: would it make more sense to embed the schema instead?
When a referenced type is actually embedded in the message, where it is being referenced, we could use the schema directly - that would make it self-contained, avoid duplicate naming and wild parent_parent_parent_* naming schemes.

So this:
```proto
message Message {
    message SubMessage {
        int64 id = 1;
        string label = 2;
    }
    string message_id = 1;
    SubMessage sub_message1 = 2;
    SubMessage sub_message2 = 3;
}
```

Would result in:
```yaml
components:
    schemas:
        Message:
            properties:
                messageId:
                    type: string
                subMessage1:
                    type: object
                    properties:
                        id:
                            type: integer
                            format: int64
                        label:
                            type: string
                subMessage2:
                    type: object
                    properties:
                        id:
                            type: integer
                            format: int64
                        label:
                            type: string
```

Instead of what it currently results in:
```yaml
components:
    schemas:
        Message:
            properties:
                messageId:
                    type: string
                subMessage1:
                    $ref: '#/components/schemas/SubMessage'
                subMessage2:
                    $ref: '#/components/schemas/SubMessage'
        SubMessage:
            properties:
                id:
                    type: integer
                    format: int64
                label:
                    type: string
```

I think that it's fair to argue, that the embedded message is probably embedded to avoid collisions and that "inlining" doesn't detract any meaning or functionality.

timburks:
@morphar Regarding inlining the nested message: I think that makes sense when we look at the spec in isolation or use it for schema validation, but for the main use case of Protocol Buffers and Discovery documents (code generation), it leaves us with no guidance about what to name a generated struct that would correspond to the nested message. Nameless nested schemas are valid in OpenAPI, so a code generator might need to accept them, but I think keeping names in our generated OpenAPI specs helps code generators produce more usable code.

I looked at what the Go Protobuf generator produces for the two [nested messages in the cloud/translate/v3 API](https://github.com/googleapis/googleapis/blob/02df998e40733c077aa0fc3f35a6b2d48aa8bf84/google/cloud/translate/v3/translation_service.proto#L977): I found structs named [`Glossary_LanguageCodePair`](https://github.com/googleapis/go-genproto/blob/81c1377c94b1698bef1982dc9e18f3260cfcbf1a/googleapis/cloud/translate/v3/translation_service.pb.go#L3757) and [`Glossary_LanguageCodesSet`](https://github.com/googleapis/go-genproto/blob/81c1377c94b1698bef1982dc9e18f3260cfcbf1a/googleapis/cloud/translate/v3/translation_service.pb.go#L3817). For the sake of generated code quality, I think this is probably the right pattern to follow for our generated OpenAPI.

timburks (approved):
@morphar Thanks again for this! You've really moved this generator forward!

morphar:
@timburks you're so very welcome :) I'm just happy to be able to contribute to a really useful project.
I've tried to push code generation in all projects I've been involved in for many, many years. And projects like this, is what is needed to make it consistent and easy to have one source of truth and save months and months of work on SDKs, documentation and "trivial" machine-to-machine communication... So again: really happy to contribute :)

morphar:
> @morphar Regarding inlining the nested message: I think that makes sense when we look at the spec in isolation or use it for schema validation, but for the main use case of Protocol Buffers and Discovery documents (code generation), it leaves us with less guidance about what to name a generated struct that would correspond to the nested message (though I suppose we could use the field name). Nameless nested schemas are valid in OpenAPI, so a code generator might need to accept them, but I think keeping names in our generated OpenAPI specs helps code generators produce more usable code.
> 
> I looked at what the Go Protobuf generator produces for the two [nested messages in the cloud/translate/v3 API](https://github.com/googleapis/googleapis/blob/02df998e40733c077aa0fc3f35a6b2d48aa8bf84/google/cloud/translate/v3/translation_service.proto#L977): I found structs named [`Glossary_LanguageCodePair`](https://github.com/googleapis/go-genproto/blob/81c1377c94b1698bef1982dc9e18f3260cfcbf1a/googleapis/cloud/translate/v3/translation_service.pb.go#L3757) and [`Glossary_LanguageCodesSet`](https://github.com/googleapis/go-genproto/blob/81c1377c94b1698bef1982dc9e18f3260cfcbf1a/googleapis/cloud/translate/v3/translation_service.pb.go#L3817). For the sake of generated code quality, I think this is probably the right pattern to follow for our generated OpenAPI.

This makes sense 👍 Let's take that path then.
I will make another pull request real soon, as I discovered that tags are being created for services without http annotations.
I will try to get the nested messages code done as well.

PR #272, opened 2021-11-28
Handle nested messages and ignore services without http annotations

morphar:


morphar:
@timburks it was pretty simple and easy to integrate your from the [gist](https://gist.github.com/timburks/44f3671a7e80b0cd7a9b922192f3941b) 👍

morphar:
By the way... I'm creating a protoc-gen-jsonschema, which I think would fit perfectly into this project - do you agree or is out of scope for gnostic?

I want it to output 1 file per proto message and then use full URL references, based on a root URL given as protoc param.
I found [chrusty/protoc-gen-jsonschema](https://github.com/chrusty/protoc-gen-jsonschema), but I'm not entirely happy with the output, which uses the proto package as prefix for references and puts everything in `definitions`.
I'm not against it as such, I just can't make it generate the output that I need.

timburks:
@morphar thanks for this! For `protoc-gen-jsonschema`, I think that fits if you'd like to add it. The OpenAPI generator seems like a clearer fit because it uses gnostic structures to build the OpenAPI description. We currently don't have a set of generated protos for JSON Schema, so I'm wondering how you'd like to generate JSON Schema (direct to JSON?). I think it would be good to do more with JSON Schema - an earlier draft of JSON Schema is at the core of the project (the `gnostic-generator` generates all of the proto models from JSON schemas). Feel free to start a PR if you'd like.

apps/protoc-gen-openapi/generator/openapi-v3.go
```diff
@@ -225,11 +231,25 @@ func (g *OpenAPIv3Generator) formatMessageRef(name string) string {
 }
 
 func (g *OpenAPIv3Generator) formatMessageName(message *protogen.Message) string {
+
```
timburks, line 234: The name generation code (below) is essentially in here twice - I lazily copied-and-pasted it between this function and `fullMessageTypeName` in the gist. I think it would be good (but optional  in this PR) to refactor into a common function.

timburks (approved):

apps/protoc-gen-openapi/generator/openapi-v3.go
morphar: You're are right - I guess I got lazy too as it just worked :)
I have moved the code to a function.

morphar:
I can see what you mean by OpenAPI being a clearer path as JSON Schemas is not directly related/needed for any of the documentation and conversion being done by gnostic.

The reason I need it, is that I am dynamically building database queries that utilizes a JSON Schema - the schemas can change at any time, but I also have some models defined in protobuf, that should be taken into consideration.
And that is why I would like to have those defined in JSON Schema as well.

It might be such an outlier case, that it might be a perfect fit in my case, but not in most other cases?

timburks (approved):
@morphar Thanks for making this additional change! I'll merge this now.

timburks:
@morphar The JSON schema generator SGTM - I would be happy to help with it also. It reminds me of an undeveloped idea from early in the gnostic project - originally I used the OpenAPI v2 JSON schema as a source for generating a protobuf representation and a compiler for OpenAPI. The only reason I used JSON schema was because there was that JSON schema for OpenAPI v2. But then after writing the generator (in the `gnostic-generator` directory), I realized that it could be useful for generating protobuf models of any format with a JSON schema, but without a use for a general JSON schema-to-protobuf generator, I didn't take that much further (only to generate models for OpenAPI v3 and Google's API Discovery format). What you're suggesting is basically going in the opposite direction: from protobuf to JSON schema, but in exploring the protobuf-JSON correspondence, I think it fits and could "complete the loop" by giving us a way to make bidirectional conversions between protobuf and JSON schema descriptions of messages.

morphar:
@timburks That was my exact thinking: full circle :)
I am currently building up a toolset for conversion between as many API related "things" as possible.
I find a lot of useful tools, but most of them are opinionated (bad in this context) and only goes in one direction.
That was why I thought gnostic was a very interesting project - it's has way more tools and in this context, being opinionated makes everything work nicely together.

I am currently building the protoc-gen-jsonschema app and I think it would benefit from being within the gnostic "eco-system".
I noticed that the jsonschema lib will only be able to support draft-04, draft-06 and draft-07 as the newer drafts changed to `$refs` instead of `definitions`. This should not be a big issue though, as newer drafts can easily use older draft schemas.
I just open sourced the JSON Schema lib we have build internally in FlowStack: [go-jsonschema](https://github.com/flowstack-com/go-jsonschema) (WIP), I would be happy to "donate" the project or parts of it to gnostic, if that would make sense in any way, but a lot of the code is validation code, which is probably not useful in gnostic.

Issue #275, opened 2021-12-14
 Explicit 'optional' labels are not supported


maxlein:
I can't build my swagger services because optionals are not allowed.  

`simple_rpc.proto: is a proto3 file that contains optional fields, but code generator protoc-gen-openapi hasn't been updated to support optional fields in proto3. Please ask the owner of this code generator to support proto3 optional.--openapi_out: `  

Is there any work done yet?

morphar:
Hi @maxlein, I have done some work on the protoc-gen-openapi and I may be able to add optional.
Do you have an example of the proto file that fails? Just a simple example with the optional will do.
If you also have an example of the expected Open API file output, that would also be helpful.

maxlein:
Here you go:  

```protobuf
syntax = "proto3";

import "google/api/annotations.proto";
import "google/api/client.proto";

// we build swagger config with https://github.com/google/gnostic/tree/master/apps/protoc-gen-openapi
option go_package = "./"; // only for proto swagger generator

// Interface exported by the server.
service HostIf {
  // TODO add server example host so we can use swagger to call services
  option (google.api.default_host) = "localhost"; // seems to have no effect

  // A simple RPC.

  rpc GetStation(StationName) returns (StationInfo) {
    
    option (google.api.http) = {
      get: "/v1/station"
    };
    option (google.api.method_signature) = "station";
  }

}

message StationName {
  string name = 1;
  Metadata metadata = 2;
  bool enabled = 3;
  optional uint64 dummy = 4;
}

message StationInfo {
  string name = 1;
  bool operational = 2;
}

message Metadata {
  map<string, string> data = 1;
}

```   
Generated openapi yaml ( optional was removed here ):
I don't know how an optional field should look like in json to be honest..  
Just ignoring would be an option to explicitly choose if needed, as in proto 3, I think every field is optional implicitly.  
Since ~3.17 or so you can add an optional tag, which implicitly wrap the field with a "oneof", so you can check for existance if I am right.  
  
Also it looks like all datatypes are converted to strings ...
I am using protoc version 3.17.3  

```
# Generated with protoc-gen-openapi
# https://github.com/google/gnostic/tree/master/apps/protoc-gen-openapi

openapi: 3.0.3
info:
    title: HostIf API
    description: Interface exported by the server.
    version: 0.0.1
servers:
    - url: http://127.0.0.1
paths:
    /v1/station:
        get:
            tags:
                - HostIf
            operationId: HostIf_GetStation
            parameters:
[...]

PR #276, opened 2021-12-20
Protoc gen openapi

morphar:
Fixes for #275 + additional refinement of query parameter handling.


apps/protoc-gen-openapi/main.go
```diff
@@ -20,23 +20,28 @@ import (
...
 	conf := generator.Configuration{
-		Version:     flags.String("version", "0.0.1", "version number text, e.g. 1.2.3"),
-		Title:       flags.String("title", "", "name of the API"),
-		Description: flags.String("description", "", "description of the API"),
-		Naming:      flags.String("naming", "json", `naming convention. Use "proto" for passing names directly from the proto files`),
+		Version:       flags.String("version", "0.0.1", "version number text, e.g. 1.2.3"),
+		Title:         flags.String("title", "", "name of the API"),
+		Description:   flags.String("description", "", "description of the API"),
+		Naming:        flags.String("naming", "json", `naming convention. Use "proto" for passing names directly from the proto files`),
+		CircularDepth: flags.Int("depth", 2, `depth of recursion for circular messages`),
```
timburks, line 34: Could you point me to any examples in the tests (or elsewhere) that require this? I'm wondering how someone would know what value to use for this.

morphar: I can see how that's really not obvious from a user perspective...
I just wanted to make sure, that it is configurable, since I can think of cases for wanting both a higher and a lower value.

Lower: if it's references that is only set in code and not expected to be given as query params.
Higher: if it's used for something that resembles a family tree, where the depth could be really high.

I can't seem to figure out, how to link to the patch, so I'll use links to the file in the pull requests.

I start a map containing visit depths here:
https://github.com/google/gnostic/blob/7aa98fc70ada0b953b2698adad65480615933359/apps/protoc-gen-openapi/generator/openapi-v3.go#L366-L372

The map is being used here:
https://github.com/google/gnostic/blob/7aa98fc70ada0b953b2698adad65480615933359/apps/protoc-gen-openapi/generator/openapi-v3.go#L388-L407

With a check on the CircularDepth var here:
https://github.com/google/gnostic/blob/7aa98fc70ada0b953b2698adad65480615933359/apps/protoc-gen-openapi/generator/openapi-v3.go#L397

I have added recursive messages in the protobuftypes test:
https://github.com/google/gnostic/blob/7aa98fc70ada0b953b2698adad65480615933359/apps/protoc-gen-openapi/examples/tests/protobuftypes/message.proto#L48-L55

and added them to Message (line 71 and 76):
https://github.com/google/gnostic/blob/7aa98fc70ada0b953b2698adad65480615933359/apps/protoc-gen-openapi/examples/tests/protobuftypes/message.proto#L67-L76

And the test .json files has been updated to reflect the recursion, e.g.:
https://github.com/google/gnostic/blob/7aa98fc70ada0b953b2698adad65480615933359/apps/protoc-gen-openapi/examples/tests/protobuftypes/openapi.yaml#L26-L45

Currently, only the default (2) is being tested though.

morphar:
@timburks I just merged everything together 👍

timburks:
@morphar Thanks for these updates - when I ran the latest protoc-gen-openapi from this branch on an API with field masks, I noticed the following change:
```
-                - name: updateMask
+                - name: updateMask.paths
```
This is inconsistent with the [documented JSON encoding for field masks](https://github.com/protocolbuffers/protobuf/blob/01e84b129361913e5613464c857734fcfe095367/src/google/protobuf/field_mask.proto#L180), which I guess we only satisfied accidentally prior to this change. I think we can fix this by adding a new case [here](https://github.com/google/gnostic/blob/dd0d92fd224dee6096d9f9c19bbb69cb88c57b61/apps/protoc-gen-openapi/generator/openapi-v3.go#L586). I will try that now and follow up.

timburks:
@morphar The change I mentioned wasn't enough - I realize now that we need to prevent the recursive expansion of updateMask into updateMask.paths - in other words, blocking expansion of FieldMask messages. Can you suggest a way to do that?

timburks:
@morphar I think I found a way to handle field masks - see https://github.com/morphar/gnostic/pull/2

morphar:
Thank you @timburks 👍
I have reviewed, tested and merged the pull request.

timburks:
@morphar thanks! Confirmed here also. Merging.

timburks (approved):

morphar:
@timburks Nice 👍 Thank you :)

PR #281, opened 2022-01-04
protoc-gen-openapi: schema types

olekturbo:
Schema types are required when auto generating code from OpenAPI.
[Issue](https://github.com/google/gnostic/issues/263) was created some time ago.
```
Schema:
    type: object
    required:
     - name
    properties:
      name:
        type: string
        description: service name
```

### DIFF
--- a/apps/protoc-gen-openapi/examples/google/example/library/v1/openapi.yaml
+++ b/apps/protoc-gen-openapi/examples/google/example/library/v1/openapi.yaml
@@ -314,6 +314,7 @@ paths:
 components:
     schemas:
         Book:
+            type: object
             required:
                 - name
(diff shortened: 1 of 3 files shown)


morphar:
Hi @olekturbo,

I agree that it should not be a problem to add the type for all schemas.
It would actually make it more correct.

This pull request will break all tests that are not within the google example, so I took the liberty to create a new pull request (#282) with the same one-line change, but with all tests modified to reflect the change.
Hope that's o.k. with you :)

olekturbo:
Thats ok

timburks:
@olekturbo Thank you! Closing this since #282 was merged.

PR #301, opened 2022-01-23
[+] wrapped allof for $ref message

ppaanngggg:
wrap `$ref` message with allof:

1. keep the `json` same;
2. `writeonly` and `readonly` work now for the whole message;

refer to https://stackoverflow.com/questions/51402156/how-to-declare-a-ref-property-as-readonly-in-openapi-swagger


morphar:
LGTM 👍 
Does this pass tests?

ppaanngggg:
Can't pass the tests.

Because there are `allOf` wappers of schema, for example:

After:
image

Before:
image

morphar:
I don't think they are 100% in sync, but the closer they are, the better.

ppaanngggg:
It seems like openapi 3.1 support `readOnly` for `$ref`, so maybe upgrade to 3.1 is best way to solve this problem.

#307

morphar:
There has been some talks regarding upgrading from OpenAPI version from 3.0 to 3.1.
I'm not sure what is the best fit for gnostic:
1. Only support OpenAPI 3.1
2. Support both OpenAPI 3.0 and OpenAPI 3.1

As far as I can see, the major difference is the full compatibility with JSON Schema in OpenAPI 3.1.

@timburks what are your thoughts on this? And what makes most sense regarding the future of gnostic? :)

ppaanngggg:
> 1. Only support OpenAPI 3.1

I prefer `1. Only support OpenAPI 3.1`.
Because new is always better.

ppaanngggg:
I can do better.

I will add an option to control this feature, then I can add a single test for this feature and others can disable this feature by keeping default.

PR #307, opened 2022-02-13
cmd/protoc-gen-openapi/generator: fix Openapi version to 3.1.0

zchee:
cmd/protoc-gen-openapi/generator: fix Openapi version to 3.1.0.

github.com/google/gnostic/openapiv3 already supports 3.1.0.


zchee:
cc: @timburks

morphar:
Thank you @zchee, but the supported version is 3.0.x, not 3.1.x.
3.1.x supports a full JSON Schema spec, while 3.0.x doesn't.
Some of the generated parts of the OpenAPI .yaml file was designed to support 3.0.x.

Changing to 3.1.0 would cause 3.0.x consumers to fail, while 3.1.x consumers can probably consume 3.0.x.

PR #314, opened 2022-03-01
Support $ref alongside description in OpenAPI V3

Jefftree:
Copied from the latest OpenAPI 3.1 spec: https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.json#L1072-L1081

Add support for $ref to have a description alongside it in the same object in the proto model for OpenAPI V3.0 and V3.1. This is an expected property for 3.1 and a relaxed constraint for 3.0.

/assign @timburks


openapiv3/OpenAPIv3.proto
```diff
@@ -208,8 +208,8 @@ message Info {
   Contact contact = 4;
   License license = 5;
   string version = 6;
-  repeated NamedAny specification_extension = 7;
-  string summary = 8;
+  string summary = 7;
```
Jefftree, line 211: This change was generated automatically after running the regeneration script. Is this expected?

timburks: I've been manually changing this back to maintain backwards-compatibility of the protos. Could you do that here? Essentially I'm asking for the summary tag to be 7 and specification_extension should be 8.

timburks: That would also require regenerating OpenAPIv3.pb.go

Jefftree: Fixed. Updated both the .proto and .pb.go file.

timburks (approved):
@Jefftree Thanks for these changes and updates! LGTM, will merge

PR #316, opened 2022-03-14
Change CMD directory to use its own go.mod

Jefftree:
Prevent dependencies only used for cmd from being included in imports.

Ref: 
https://github.com/kubernetes/kubernetes/pull/108644#discussion_r826355925
https://github.com/google/cadvisor/pull/2437


timburks:
Thanks @Jefftree, see #317 for some extended thoughts on this.

Jefftree:
Closing in favor of #317

PR #318, opened 2022-03-16
fix ref conflict

### LINKED ISSUE #309: Supporting duplicate models (different packages) in protoc-gen-openapi
I have a setup where the same model name is used in different packages (or different versions of the same package). In my protos, these are in separate directories (/foo/v1/bar.proto and /foo/v2/bar.proto, both declaring Bar).

The openapi (v2) generator from grpc-gateway handles this by including the version or if needed full package name in the generated model types, whereas with gnostic, I get duplicate types and errors in later tooling using the openapi spec.

What I would ideally want is to get the same style output in JavaScript based on the openapi spec as I do with other tooling directly using the grpc/proto definitions, with models keeping their names but packaged up in directories matching the proto package / directory structure, though I'm not sure if there's a way of representing this in openapi.

Is there a way of either getting the gnostic protoc-gen-openapi tool to work with this sort of setup, or alternatively, best practices for grpc/proto that means that this would not be a problem?

tonybase:
Fixed the same schema name in different packages:
```
$ref: '#/components/schemas/Book'
->
$ref: '#/components/schemas/.library.v1.Book'
```
Example:
```
# Generated with protoc-gen-openapi
# https://github.com/google/gnostic/tree/master/cmd/protoc-gen-openapi

openapi: 3.0.3
info:
    title: Greeter API
    description: The greeting service definition.
    version: 0.0.1
paths:
    /helloworld/{name}:
        get:
            tags:
                - Greeter
            description: Sends a greeting
            operationId: Greeter_SayHello
            parameters:
                - name: name
                  in: path
                  required: true
                  schema:
                    type: string
            responses:
                "200":
                    description: OK
                    content:
                        application/json:
                            schema:
                                $ref: '#/components/schemas/.helloworld.v1.HelloReply'
components:
    schemas:
        .helloworld.v1.HelloReply:
            type: object
            properties:
                message:
                    type: string
            description: The response message containing the greetings
tags:
    - name: Greeter
```

fix #309

### DIFF
--- a/cmd/protoc-gen-openapi/generator/openapi-v3.go
+++ b/cmd/protoc-gen-openapi/generator/openapi-v3.go
@@ -317,17 +317,7 @@ func getMessageName(message protoreflect.MessageDescriptor) string {
 }
 
 func (g *OpenAPIv3Generator) formatMessageName(message *protogen.Message) string {
-	name := getMessageName(message.Desc)
-
-	if *g.conf.Naming == "proto" {
-		return name
-	}
-
-	if len(name) > 0 {
-		return strings.ToUpper(name[0:1]) + name[1:]
-	}
-
-	return name
+	return fullMessageTypeName(message.Desc)
 }
 
 func (g *OpenAPIv3Generator) formatFieldName(field *protogen.Field) string {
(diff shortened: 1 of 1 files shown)


jeffsawatzky:
@timburks this can be closed due to #324

timburks:
@jeffsawatzky @tonybase thanks! Closing this now based on @jeffsawatzky's comment.

PR #319, opened 2022-03-16
Use correct RFC 3339 format as defined in OpenAPI spec

adolfo:
"RFC3339" is not a valid format for an OpenAPI date-time notation. The correct format is "date-time" which is a string that conforms to RFC 3339.

> OpenAPI defines the following built-in string formats:
> ...
> **date-time** – the date-time notation as defined by RFC 3339, section 5.6, for example, 2017-07-21T17:32:28Z
> ...

[source 1](https://swagger.io/docs/specification/data-models/data-types/#string)
[source 2](https://github.com/google/gnostic/blob/ed308846fc77db7667936d27d7ac1fb278b74216/openapiv3/schema-generator/3.0.2.md#format)


timburks (approved):

timburks:
Fixed in #330, thanks @adolfo and @jeffsawatzky !

PR #324, opened 2022-03-28
[protoc-gen-openapi] Support Fully Qualified Schema Names and Annotations for File/Document and Method/Operation

### LINKED ISSUE #308: Marking a service or method as requiring bearer authentication in proto/grpc definition
I'm comparing the grcp-gateway protoc plugin for openapi (v2) with the one from here for openapi v3. While the built in functionality seems to "just work", I'm not sure how to customise the openapi output further, such as for specifying the security schemes.

I have seen examples of this in the grpc-gateway plugin using custom annotations, e.g. https://github.com/grpc-ecosystem/grpc-gateway/issues/1089

Is there something similar for using the gnostic openapi codegen plugin?

### LINKED ISSUE #309: Supporting duplicate models (different packages) in protoc-gen-openapi

### LINKED ISSUE #322: openapi_out: How to overwrite operationId ?
I am trying to figure out, how I could overwrite the operationId with gnostic.  
Because in the documentation it looks bad if every endpoint is prefixed with the service name.  

I saw different proto annotation files for swagger and openapiv2, but nothing for v3.  

I just tried to get it working by using it like this:  

```
syntax = "proto3";

import "google/api/annotations.proto";
import "google/protobuf/empty.proto";
import "protoc-gen-openapiv2/options/annotations.proto";

// we build swagger config with https://github.com/google/gnostic/tree/master/apps/protoc-gen-openapi
option go_package = "./"; // only for proto swagger generator

service TaskService {
  
  rpc Execute(TaskData) returns (TaskResponse) 
  {
      option (google.api.http) = {
        post: "/v1/execute"
        body: "*" // puts all message fields in body
      };
      option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
        operation_id: "OtherId";
      };
  }
}
```  

But I get no error and also no "OtherId" as operationId.  
OperationId is `TaskService_Execute`  

Should this be working with v3, or am I mixing up v2 with v3?

jeffsawatzky:
This does a couple things:
1) Refactor the code a little to provide the actual message and field instances to more places instead of just the descriptor
2) Removed the `formatMessageRef` function as it was basically a duplicate of `formatMessageName`, and with the above refactor it wasn't needed anymore
3) Added support for adding the message package name to the schema name to support fully qualified schema names. This is done with the new option `fq_schema_naming`
4) Added support for options on the File (which are used on the OpenAPI Document) and Method (for the OpenAPI Operation)
4) Added some tests for the new features
5) Added functionality to easily generate new test fixtures and update existing test fixtures if needed.

Fixes #308 
Fixes #309
Fixes #322


jeffsawatzky:
@googlebot I signed it!

jeffsawatzky:
@timburks Could you have a look at the PR when you have time. Thanks!

jeffsawatzky:
@timburks I'm in the process of bringing the functionality of protoc-gen-openapi close to the grpc-gateway-openapiv2 one in terms of customization, but all may changes are based on each other.

Do you want everything in one huge PR? Or would you prefer smaller ones? If smaller ones, then this PR is what most of the newer ones will be based off of so I would really appreciate it if the could get reviewed, and if you approve, merged.

Thanks!

timburks:
@jeffsawatzky Thanks for this work and your explanations. I'll review (and expect to approve) this today; from your note at the top, it all SGTM.

Would you mind spelling out what you mean by "bringing the functionality of protoc-gen-openapi close to the grpc-gateway-openapiv2 one in terms of customization"? What specific things are you hoping to add? Just wanting to sync on top-level goals.

Thanks!

/cc @morphar

jeffsawatzky:
@timburks being able to embed OpenAPI elements into the proto as options so you can define extra things like security on a method within the proto file. I've added examples of this to the tests in this PR.

Also, envoy and grpc-gateway return errors using the `google.rpc.Status` message type and currently this plugin doesn't support the `google.protobuf.Any` message properly which is used by `google.rpc.Status` message (see #327 which adds support for this)

openapiv3/annotations.proto
```diff
@@ -0,0 +1,52 @@
...
+// is to use an abbreviation of the package name. Something short, but
+// hopefully unique enough to not conflict with things that may come along in
+// the future. 'GPB' is reserved for the protocol buffer implementation itself.
+option objc_class_prefix = "OAS";
+
+// The Go package name.
+option go_package = "./openapiv3;openapi_v3";
+
+extend google.protobuf.FileOptions {
+  Document document = 1042;
```
timburks, line 47: So (just digesting), this allows an arbitrary OpenAPI document to be attached to a file, presumably a partial document that gets overlaid/merged onto the generated OpenAPI spec.

openapiv3/annotations.proto
```diff
@@ -0,0 +1,52 @@
...
+}
+
+extend google.protobuf.MethodOptions {
+  Operation operation = 1042;
```
timburks, line 51: Similarly, this merges an Operation into the operation generated for a method. This is a gnostic type (as is Document) - and I notice from your example that you're assigning them with valid structs, which are unintuitive to most OpenAPI users but fit the gnostic protobuf model - and I think that consistency is good. Without looking through your generator code (yet), I guess that makes it easier to merge these with the generated doc since that's working on gnostic protobuf-based structs.

cmd/protoc-gen-openapi/generator/openapi-v3.go
```diff
@@ -103,6 +104,12 @@ func (g *OpenAPIv3Generator) buildDocumentV3() *v3.Document {
 
 	for _, file := range g.plugin.Files {
 		if file.Generate {
+			// Merge any `Document` annotations with the current
+			extDocument := proto.GetExtension(file.Desc.Options(), v3.E_Document)
+			if extDocument != nil {
+				proto.Merge(d, extDocument.(*v3.Document))
```
timburks, line 110: Nice, this seems really clever and elegant. I wonder if there are any other places (besides methods and files) where it would make sense to attach a partial document - maybe messages or fields?

cmd/protoc-gen-openapi/main.go
```diff
@@ -27,20 +27,21 @@ var flags flag.FlagSet
...
-		Title:         flags.String("title", "", "name of the API"),
-		Description:   flags.String("description", "", "description of the API"),
-		Naming:        flags.String("naming", "json", `naming convention. Use "proto" for passing names directly from the proto files`),
-		EnumType:      flags.String("enum_type", "integer", `type for enum serialization. Use "string" for string-based serialization`),
-		CircularDepth: flags.Int("depth", 2, "depth of recursion for circular messages"),
+		Version:        flags.String("version", "0.0.1", "version number text, e.g. 1.2.3"),
+		Title:          flags.String("title", "", "name of the API"),
+		Description:    flags.String("description", "", "description of the API"),
+		Naming:         flags.String("naming", "json", `naming convention. Use "proto" for passing names directly from the proto files`),
+		FQSchemaNaming: flags.Bool("fq_schema_naming", false, `schema naming convention. If "true" prefixes the schema name with the proto message package name`),
```
timburks, line 34: Let's spell out what "fq" stands for here in the help string, maybe like this:
```
If "true", generates fully-qualified schema names by prefixing them with the proto message package name.
```

openapiv3/annotations.proto
```diff
@@ -0,0 +1,52 @@
...
+// It should at a minimum be 3 characters long, all uppercase, and convention
+// is to use an abbreviation of the package name. Something short, but
+// hopefully unique enough to not conflict with things that may come along in
+// the future. 'GPB' is reserved for the protocol buffer implementation itself.
+option objc_class_prefix = "OAS";
+
+// The Go package name.
+option go_package = "./openapiv3;openapi_v3";
+
+extend google.protobuf.FileOptions {
```
timburks, line 46: The other files in this directory are automatically-generated - that's not a problem, and I think this is a good place to put this file, especially since it refers to types defined inside the package.

Eventually (not necessarily in this PR) I think it would be good to have some documentation here of how these annotations work. [google/api/http.proto](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto) is a good example of this - though I'm not thinking of anything nearly as elaborate as that, that file seems to be a good example of where to put proto documentation -- in comments in the protos so that it never gets separated.

timburks (approved):
@jeffsawatzky thanks, this looks great. I really like the options annotations and how well they fit into the gnostic data structures.

I'll let this sit overnight (California time) in case @morphar has time to look and comment and if no objections come up, I think we can merge this tomorrow morning (PDT).

morphar:
Sorry, I was away from my computer, I'll have a look at it now :)

morphar:
I have no comments to add whatsoever! It looks really good to me 👍

As far as I can tell, all existing things just works, while fixing Any and adding some nice functionality.
I just did a full run of tests on a local project (gRPC, gRPC-Web, Envoy, OpenAPI, etc.) and nothing breaks there either.

Really nice @jeffsawatzky 👍

openapiv3/annotations.proto
jeffsawatzky: @timburks Correct. It will do a [proto merge](https://pkg.go.dev/github.com/golang/protobuf/proto#Merge) on the result `Document` and any `openapi.v3.document` options it finds while compiling the proto files.

openapiv3/annotations.proto
jeffsawatzky: Protocol Buffer options need to be defined using protocol buffers themselves. So you will never get a 1-1 mapping of the options structure with OpenAPI structure as protocol buffers are strictly typed and yaml/json are not. Also, this is similar to how the [protoc-gen-openapiv2](https://github.com/grpc-ecosystem/grpc-gateway/tree/master/protoc-gen-openapiv2) generator works from the grpc-gateway project (though they use protocol buffers that are specific to OpenAPI v2/Swagger). I think most developers that will use this will understand this.

Since I need to use protocol buffer messages for the options, I reused the same ones that gnostic was already using to make merging them easier, as you said.

cmd/protoc-gen-openapi/generator/openapi-v3.go
jeffsawatzky: Yeah, I can add the same for Messages=>Schemas and Fields=>Parameters. But didn't want to do too much in case you reject the whole idea. I can do this in a separate PR.

ppaanngggg:
so, merge this plz. I really need this feature too. 👍

cmd/protoc-gen-openapi/main.go
jeffsawatzky: I have a couple more PRs coming. I will update the help string there, if that's ok with you.

jeffsawatzky:
@timburks / @morphar once you merge this PR, can you also take a look at my other PR #327 which is based on this? I have some more functionality that I would like to add for my use case, such as adding an option for a `default` response for error responses that use the `google.rpc.Status` message (since Envoy and grpc-gateway both use `google.rpc.Status` as their error response). But it would be helpful if these two PRs could be merged first.

cmd/protoc-gen-openapi/main.go
timburks: Yes, SGTM

cmd/protoc-gen-openapi/generator/openapi-v3.go
timburks: I think that would be great, thanks!

ppaanngggg:
Could you please add a Tag on this commit? @timburks @morphar

morphar:
It's only @timburks who can add tags and versions. I'm just a contributor :)

PR #325, opened 2022-03-29
Feat/fullpackage

ppaanngggg:
If two package have the same name message, openapi will go wrong.

So I add an option to output the full name of service and message to avoid same name error.


jeffsawatzky:
I did something similar in #324 😀 

I also added annotations so you can customize the output a bit more.

ppaanngggg:
> I did something similar in #324 😀
> 
> I also added annotations so you can customize the output a bit more.

Great, I will waiting for your PR to be merged

PR #327, opened 2022-03-29
[protoc-gen-openapi] Support google.protobuf.Any and google.rpc.Status

jeffsawatzky:
Note that this PR is based off of my previous PR #324 which is not yet merged. So while the changes here look big, here are the actual changes relevant to this PR:
https://github.com/jeffsawatzky/gnostic/compare/fq-names...jeffsawatzky:grpc-status

# Purpose
Envoy and grpc-gateway use the `google.rpc.Status` message to report error messages back as can be seen here:
https://github.com/envoyproxy/envoy/blob/9dd8ae3c69a9917d97a886ed03e1c010dcd9b098/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L827L860

`google.rpc.Status` provides a way to report additional error details in the `details` property, which is a list of `google.protobuf.Any` which has special JSON encoding as can be seen here:
https://developers.google.com/protocol-buffers/docs/proto3#json

Note that `google.protobuf.Any` and `google.protobuf.Value` are different. `google.protobuf.Value` can be any JSON value, whereas `google.protobuf.Any` can be any protobuf message type and has a special `@type` property when transcoded so that you can determine which protobuf message definition to use.

We don't have control over the value of `@type` and it will always be the fully qualified name of the protobuf message. So if people what to be able to use this to map back to an OpenAPI schema definition then they will need to use fully qualified names (which I added in PR #324)


jeffsawatzky:
@timburks here is another one for you

timburks:
@jeffsawatzky could you merge main (now containing #324) and resolve any conflicts? I'm not expecting any problems, but that would clarify what's new here.

jeffsawatzky:
@timburks done (rebased everything)

cmd/protoc-gen-openapi/examples/tests/protobuftypes/openapi.yaml
```diff
@@ -246,7 +246,7 @@ components:
                 - type: array
                   items:
                     $ref: '#/components/schemas/AnyJSONValue'
```
timburks, line 248: On my first read-through, I'm wondering why this didn't change to google.protobuf.Value

cmd/protoc-gen-openapi/examples/tests/protobuftypes/openapi_json.yaml
```diff
@@ -246,7 +246,7 @@ components:
                 - type: array
                   items:
                     $ref: '#/components/schemas/AnyJSONValue'
```
timburks, line 248: also here, why isn't this google.protobuf.Value?

timburks (commented):
Just a few initial questions while I'm reading this...

cmd/protoc-gen-openapi/examples/tests/protobuftypes/openapi.yaml
jeffsawatzky: This file was generated with `fq_schema_naming=false` (the default) so it uses the older `AnyJSONValue` schema name for backwards compatibility.

jeffsawatzky: There was a bug in my previous PR with `google.protobuf.Value` not getting converted to the fully qualified name, which I didn't catch until I added support for `google.protobuf.Any`. This is why you see the schema name change in the `openapi_fq_schema_naming.yaml` file (which uses fully qualified naming) since I have fixed the issue in this PR.

cmd/protoc-gen-openapi/generator/openapi-v3.go
```diff
@@ -315,7 +316,16 @@ func getMessageName(message protoreflect.MessageDescriptor) string {
 }
 
 func (g *OpenAPIv3Generator) formatMessageName(message *protogen.Message) string {
+	typeName := fullMessageTypeName(message.Desc)
+
 	name := getMessageName(message.Desc)
+	if !*g.conf.FQSchemaNaming {
```
morphar, line 322: This is also a question of: should we be consistent or backwards compatible.
I would personally prefer that types are the same, regardless of the naming.
I would be surprised if I changed the naming option and got different types.

morphar (commented):
Just a thought regarding the description of `google.protobuf.Value`:
Isn't the description a bit wrong? The value can never be `null`, right?
OpenAPI (v3.0.x) doesn't support `null` as type and the schema says `oneOf`, so it **must** be one of the types - therefore it can not be null.

It's been to long, since I worked with this code, so I can't remember.

morphar:
> Isn't the description a bit wrong? The value can never be `null`, right?
I just checked a whole bunch of scenarios and validators.
It *can not* be `null`, but it's o.k. to not set the value, as long as it is not required.
So only the description is a bit wrong 👍

cmd/protoc-gen-openapi/generator/openapi-v3.go
jeffsawatzky: I'm not sure about the history of the naming here, but without it being fully qualified with the package prefix then the name of the schema would just be `Value` which seems a little too general perhaps. Same with `Any`. I am happy to switch them both to just be `Value` and `Any` in the case non-fully-qualified names if that is what you prefer. I personally will always use the fully qualified names so it doesn't matter to me. :)

morphar: Aahh... Good point! You're absolutely right 👍 It would easily risk conflicting with a user's naming, which is why the  `AnyJSONValue` name was chosen in the first place, if I remember correctly.

cmd/protoc-gen-openapi/examples/tests/protobuftypes/openapi.yaml
```diff
@@ -246,7 +246,7 @@ components:
...
-            description: AnyJSONValue is a "catch all" type that can hold any JSON value, except null as this is not allowed in OpenAPI
+            description: '`Value` represents a dynamically typed value which can be either null, a number, a string, a boolean, a recursive struct value, or a list of values. A producer of value is expected to set one of that variants, absence of any variant indicates an error. The JSON representation for `Value` is JSON value.'
```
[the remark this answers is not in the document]
jeffsawatzky: This description is coming straight off of the google.protobuf.Value message here:
https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto#L56L61
Which followed the pattern of using the message comments for the schema descriptions.

I can switch it back to a hard coded value though if that is preferred.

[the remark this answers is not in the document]
morphar: When running the schema through a JSON validator, you will get an error, if the value is `null`, but not if the property is not set at all.
So basically: it can not be null - it is our definition that is wrong, since it does not have `null` as a possible type, but we can not have `null` as a type, because OpenAPI v3.0.x does not allow that.
It will be doable in OpenAPI v3.1.0 as full JSON Schema is supported.

@jeffsawatzky and @timburks what is your opinions here?

cmd/protoc-gen-openapi/generator/openapi-v3.go
morphar: Would it make more sense to have `GoogleProtobufValue` instead of `AnyJSONValue`?

jeffsawatzky: If you are ok with a possibly breaking change, that is fine with me. I'll do what the repo owners here want. :)

cmd/protoc-gen-openapi/examples/tests/protobuftypes/openapi.yaml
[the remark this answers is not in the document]
jeffsawatzky: I'm not actually a big fan of the comment for the `Any` message...it's way too long. So I might hard code that to something smaller anyway. If you want me to change the comment here for `Value` too I can do that.

cmd/protoc-gen-openapi/generator/openapi-v3.go
morphar: :) It makes sense to me, as one of the overall goals is to be able to convert both ways, so to have the "full" name, is probably more sane, than  `AnyJSONValue`. What are your thoughts @timburks?

cmd/protoc-gen-openapi/examples/tests/protobuftypes/openapi.yaml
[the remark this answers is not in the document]
morphar: As it is actually incorrect, due to the need of breaking that definition, I think it would make sense to use something like the original `AnyJSONValue is a "catch all" type that can hold any JSON value, except null as this is not allowed in OpenAPI`.
Maybe borrow some from the protobuf comment:
`Value represents a dynamically typed value which can be any JSON value, except null`.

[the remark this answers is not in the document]
morphar: For both `AnyJSONValue` and `Value`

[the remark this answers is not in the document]
jeffsawatzky: @morphar I was referring to this big long comment for the `Any` message. It's way too huge.
https://github.com/google/gnostic/pull/327/files#diff-190e6a851c89c364cd83e2bd05fc44555cdb0d940341e5c531a4eae9c0c66308R29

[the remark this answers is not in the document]
morphar: Aaahh... Gotcha :) 👍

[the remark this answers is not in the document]
jeffsawatzky: Also, while the "null" type is not allowed, we could make the schema [nullable](https://swagger.io/specification/#fixed-fields-20) which would allow null values.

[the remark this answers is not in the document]
morphar: Nice catch!
That would also allow for proper handling of `google.protobuf.Value`, as it can then also be set to `required`:
`... absence of any variant indicates an error ...` 👍

jeffsawatzky:
@morphar I updated the name and descriptions of the `Value` and `Any` messages. I also added `nullable` to the `Value` so that it should allow null values now. This will only work with OpenAPI 3.0.x though as 3.1 removed `nullable` in favour of the `null` type. Though the version we specify is 3.0.3 so I'm not sure if this is a big deal.

cmd/protoc-gen-openapi/generator/openapi-v3.go
```diff
@@ -951,9 +951,9 @@ func (g *OpenAPIv3Generator) addSchemasToDocumentV3(d *v3.Document, messages []*
 					Value: &v3.SchemaOrReference{
 						Oneof: &v3.SchemaOrReference_Schema{
 							Schema: &v3.Schema{
-								Description: messageDescription,
+								Description: "Represents a dynamically typed value which can be either null, a number, a string, a boolean, a recursive struct value, or a list of values.",
+								Nullable:    true, // this only works with OpenAPI 3.0.x
```
morphar, line 955: Is it possible to set this to `required` as well?
Then it would adhere to both OpenAPI v3.0.x and the protobuf description, right?

morphar (commented):
Only the 1 comment. Otherwise it looks good to me 👍

morphar:
> @morphar I updated the name and descriptions of the `Value` and `Any` messages. I also added `nullable` to the `Value` so that it should allow null values now. This will only work with OpenAPI 3.0.x though as 3.1 removed `nullable` in favour of the `null` type. Though the version we specify is 3.0.3 so I'm not sure if this is a big deal.

I agree. We define 3.0.3, so we should adhere to that spec, ignoring 3.1.x in this code.
It would be nice to support v3.1.x as well though, but that's a completely different topic, how to actually handle that, I think :)

jeffsawatzky:
I don't think I need to add "required" to the `Value` schema as the "Oneofs" handle that (the value HAS to be one of the defined types, or null since we allow nullable).
Reading up on the Value type [here](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Value) I noticed that the "struct" or object type has to have `Value` values, so I did add that restriction though.

morphar:
Great! 👍

Regarding `required`: you're right 👍 It would have to be the property, that defines type as `google.protobuf.Value`, that also has `required` set to true, right?
> absence of any variant indicates an error.

jeffsawatzky:
> Great! 👍
> 
> 
> 
> Regarding `required`: you're right 👍 It would have to be the property, that defines type as `google.protobuf.Value`, that also has `required` set to true, right?
> 
> > absence of any variant indicates an error.
> 
> 
> 
> 

Yes, required would have to be set on the property of whatever parent schema defines a Value property.

cmd/protoc-gen-openapi/generator/openapi-v3.go
timburks: No strong opinion, but `GoogleProtobufValue` sounds more specific. I'm fine with whatever you two work out.

morphar:
> Yes, required would have to be set on the property of whatever parent schema defines a Value property.

I'm just trying to understand if this is actually 100% compatible now.
So `google.protobuf.Value` can basically be anything, including `null`.
If a field uses this type, it MUST be set, otherwise it's an error.
All the the type constraints are now correctly converted to OpenAPI spec.

Is the `required` part also converted to OpenAPI?
Is `required` set on the property that has `type` set to `google.protobuf.Value`? Or is this not possible at all?

jeffsawatzky:
> > Yes, required would have to be set on the property of whatever parent schema defines a Value property.
> 
> I'm just trying to understand if this is actually 100% compatible now. So `google.protobuf.Value` can basically be anything, including `null`. If a field uses this type, it MUST be set, otherwise it's an error. All the the type constraints are now correctly converted to OpenAPI spec.
> 
> Is the `required` part also converted to OpenAPI? Is `required` set on the property that has `type` set to `google.protobuf.Value`? Or is this not possible at all?

@morphar  I'm not sure if it is possible or even needed to tell you the truth. From what I can tell, it isn't possible for a schema to require itself to be present in a parent schema. For example, if a `Message` has a `Value` field, it is up to the `Message` schema to specify that the `Value` field is required, and `Value` has no way of controlling this requirement itself.

With that said, I don't think it is needed anyway. According to the [Protocol Buffer Language Guide](https://developers.google.com/protocol-buffers/docs/proto3#json):
> If a value is missing in the JSON-encoded data or if its value is null, it will be interpreted as the appropriate default value
So whether the `Value` field is there and set to null, or missing, it will be treated the same way.

Furthermore, I think what we have for `Value` might actually be overkill. `google.protobuf.Value` just means that the JSON value can be any JSON value. You can achieve the same thing with a "typeless" schema
https://stackoverflow.com/a/43328994/2233608

FWIW, the protoc-gen-openapiv2 generator just uses "object"  for `Value` types.
```
"v1Message": {
  "type": "object",
  "properties": {
    "value": {
      "type": "object"
    }
  }
}
```

morphar:
@jeffsawatzky sorry, I was unclear. I meant if we should set the OpenAPI property as required.
If we could define it typeless / schemaless, that would simplify it hugely... But... As I read it, if a field (in proto) is defined as `google.protobuf.Value`, it is wrong to not set a value, though it can be `null`, it can not, not be set.
Do you agree with that understanding?
If that is true, a property in OpenAPI, should be defined as `required` AND have the schema as it is now, right?

jeffsawatzky:
@morphar I read the following:
> Absence of any variant indicates an error
As:
"If any of the variants are missing when deserializing the binary protocol buffer version of the message then it's an error."

However, when it comes to JSON I don't think it matters since the JSON transcoding will treat a missing value as an implicit null value. The docs say:
> The JSON representation for Value is JSON value

So I think we can represent `Value` as a `typeless` schema in OpenAPI which would allow any JSON value.

So, to answer your question, I don't think a property in OpenAPI needs to be defined as `required` AND have the type defined as the `Value` schema. If the property is missing the JSON transcoding will just treat it as null, which would satisfy the "NullType" variant automatically.

jeffsawatzky:
@morphar seeing as how I have a few PRs queuing up, can we keep this PR focused on the Status and Any schemas? We can discuss the Value schema in another issue perhaps?

morphar:
@jeffsawatzky Sorry, just did a test against Envoy. You're completely right, it is not required to be set, when the protobuf type is `google.protobuf.Value` 👍 So it is possible to go typeless / schemaless!

morphar (approved):
Everything looks good to me 👍 
It's fully up to you, if you want the schemaless / typeless to be in another PR :)

morphar (approved):
I thought I approved the last review :)

morphar (approved):

morphar (approved):

morphar:
Maybe @timburks needs to approve, for the merging to be unblock?

morphar (approved):
👍  Beautiful! Thank you @jeffsawatzky :)

morphar:
And sorry for the long discussion back and forth @jeffsawatzky! But nice that it resulted in a way better solution 👍

jeffsawatzky:
@morphar No worries. I'm glad we were able to work out a solution together!

timburks (approved):

timburks:
@morphar @jeffsawatzky thanks for all this! merging now

morphar:
Great! Thanks @timburks 👍

PR #329, opened 2022-04-01
Add a Default Response for Error Responses Based on google.rpc.Status

jeffsawatzky:
# Purpose
The Envoy and grpc-gateway projects automatically respond to errors using the google.rpc.Status message type. For people using either of these projects for JSON<->gRPC transcoding they may want to have a "default" response added to all of their operations that specify this.

This adds an option to automatically add a default reponse.

@timburks / @morphar


morphar:
LGTM 👍 
Really nice features you're adding @jeffsawatzky :)

cmd/protoc-gen-openapi/main.go
```diff
@@ -27,13 +27,14 @@ var flags flag.FlagSet
...
-		CircularDepth:  flags.Int("depth", 2, "depth of recursion for circular messages"),
+		Version:         flags.String("version", "0.0.1", "version number text, e.g. 1.2.3"),
+		Title:           flags.String("title", "", "name of the API"),
+		Description:     flags.String("description", "", "description of the API"),
+		Naming:          flags.String("naming", "json", `naming convention. Use "proto" for passing names directly from the proto files`),
+		FQSchemaNaming:  flags.Bool("fq_schema_naming", false, `schema naming convention. If "true", generates fully-qualified schema names by prefixing them with the proto message package name`),
+		EnumType:        flags.String("enum_type", "integer", `type for enum serialization. Use "string" for string-based serialization`),
+		CircularDepth:   flags.Int("depth", 2, "depth of recursion for circular messages"),
+		DeafultResponse: flags.Bool("default_response", false, `add default response. If "true", automatically adds a default reponse to operations which use the google.rpc.Status message. Useful if you use envoy or grpc-gateway to transcode as they use this type for their default error responses.`),
```
timburks, line 37: nit: typo

timburks:
@jeffsawatzky just to clarify, you wrote:

```The Envoy and grpc-gateway projects automatically respond to errors using the google.rpc.Status message type```

I'm assuming this requires no configuration? Since that's the default behavior for those two transcoders, I agree that we should make it the default here also -- as you've done. I'm unsure when we would ever *not* want this.

**Comment by jeffsawatzky** [comment]
> @jeffsawatzky just to clarify, you wrote:
> 
> `The Envoy and grpc-gateway projects automatically respond to errors using the google.rpc.Status message type`
> 
> I'm assuming this requires no configuration? Since that's the default behavior for those two transcoders, I agree that we should make it the default here also -- as you've done. I'm unsure when we would ever _not_ want this.

@timburks grpc-gateway uses google.rpc.Status by default unless you go out of your way to change that. Not 100% sure about envoy, but I think it is the same. Perhaps @morphar could chime in here as he seems to have more experience with envoy. But I do know this behaviour is available through envoy.

I didn't make this the default, I added an option to enable it which is disabled by default. Did this to maintain backwards compatibility. I also wasn't sure if adding envoy/grpc-gateway specific stuff should be the default behaviour.

**Comment by morphar** [comment]
My experience is limited to using Envoy in my current and previous main projects ;)

To answer your question though:
You're right. Envoy will use `google.rpc.Status` for any errors returned by a gRPC method.
I don't even know, if it's possible to change that.

We have earlier referred to these, to decide on how things should work:
- [Transcoding HTTP/JSON to gRPC](https://cloud.google.com/endpoints/docs/grpc/transcoding)
- [HTTP and gRPC Transcoding](https://google.aip.dev/127) (API Improvement Proposals)

[Errors](https://google.aip.dev/193) (API Improvement Proposals) says:
`Services **must** return a google.rpc.Status message when an API error occurs...`

So I'm assuming that we are never ever supposed to change how that works.

As I understand it, Envoy is **_the_** reference implementation for JSON transcoding gRPC.
So it should be save to assume that if Google AIP and API docs says it and Envoy does it, then it's how it should work.

**Comment by jeffsawatzky** [comment]
Ok, so it sounds like I should make this the default with an option to turn it off then?

**Comment by morphar** [comment]
In my opinion, that would definitely make most sense :)

**Comment by jeffsawatzky** [comment]
Ok, I've made the default response feature default to true by default (that's a lot of defaults...)

**Review by morphar — APPROVED** [review]
Very nice! Thanks @jeffsawatzky 👍

File: cmd/protoc-gen-openapi/generator/generator.go [hunk 2]
```diff
@@ -24,28 +24,37 @@ import (
...
-	CircularDepth  *int
-	FQSchemaNaming *bool
+	Version         *string
+	Title           *string
+	Description     *string
+	Naming          *string
+	FQSchemaNaming  *bool
+	EnumType        *string
+	CircularDepth   *int
+	DeafultResponse *bool
```
adolfo, line 45: Typo. `DefaultResponse`

timburks (approved):
@jeffsawatzky thanks for clarifying and looking into this!

PR #336, opened 2022-04-05
Feat/oneof support

ppaanngggg:
support oneof schema as protobuf jsonpb marshaler


cmd/protoc-gen-openapi/examples/tests/oneofoptions/openapi.yaml
```diff
@@ -0,0 +1,77 @@
...
+            required:
+                - plain_text
+            type: object
+            properties:
+                message_id:
+                    type: string
+                body:
+                    $ref: '#/components/schemas/Message_Body'
+                plain_text:
+                    type: string
```
timburks, line 54: Just clarifying: this change flattens the proto oneof into distinct fields for each possibility... or (looking further)

cmd/protoc-gen-openapi/examples/tests/oneofoptions/openapi_oneofwrapped.yaml
```diff
@@ -0,0 +1,85 @@
...
+                    description: The type of the serialized message.
+            additionalProperties: true
+            description: Contains an arbitrary serialized message along with a @type that describes the type of the serialized message.
+        Message:
+            type: object
+            properties:
+                messageId:
+                    type: string
+                content:
+                    oneOf:
```
timburks, line 50: with an option, we would represent proto oneofs as OpenAPI oneOfs

cmd/protoc-gen-openapi/main.go
```diff
@@ -35,6 +35,7 @@ func main() {
 		EnumType:        flags.String("enum_type", "integer", `type for enum serialization. Use "string" for string-based serialization`),
 		CircularDepth:   flags.Int("depth", 2, "depth of recursion for circular messages"),
 		DefaultResponse: flags.Bool("default_response", true, `add default response. If "true", automatically adds a default response to operations which use the google.rpc.Status message. Useful if you use envoy or grpc-gateway to transcode as they use this type for their default error responses.`),
+		OneofType:       flags.String("oneof_type", "ignore", `type for oneof wrap. Use "wrapped" to wrap fields into oneof's name property`),
```
timburks, line 38: Does "ignore" mean that we omit the oneof fields or that we flatten them (as done in the tests at the top of the PR)? If we're flattening, then I think "flattened" (or just "flat") might be a better option value.

Also, for general conciseness, how about making these values "flat" and "wrap"? I think the option name could just be "oneof".

This is a bit different from "enum_type", which is specifying an actual type (integer or string), so I don't think the "_type" suffix is necessary or clarifying here.

timburks (commented):
@ppaanngggg Thanks, this looks great! The main question that I have is "does Envoy (or grpc-gateway) do this?" Would the transcoded APIs work with clients generated from both kinds of specs? (containing flattened and wrapped oneOfs)

cmd/protoc-gen-openapi/examples/tests/oneofoptions/openapi_oneofwrapped.yaml
ppaanngggg: `represent proto oneofs as OpenAPI oneOfs`
yes, this is what I want. without it, it is not clear in api defination with oneof.

cmd/protoc-gen-openapi/main.go
ppaanngggg: great, I will adjust this option. `oneof` with available value `flat` and `wrap`, much more understandable

ppaanngggg:
In fact, I dont use envoy or grpc-gateway to do grpc-http transfer. 

But, I check grpc-gateway code, I think it can use [jsonpb](https://github.com/grpc-ecosystem/grpc-gateway/blob/e2854f5e47862dee81899502ad8a3070a701bc73/runtime/marshal_jsonpb.go) as runtime instead of [go build in json marshaler](https://github.com/grpc-ecosystem/grpc-gateway/blob/e2854f5e47862dee81899502ad8a3070a701bc73/runtime/marshal_json.go)

If you use jsonpb, you should use `wrap` mode, use buildin, you should use `flat` mode.

In envoy, I track the source code to https://github.com/grpc-ecosystem/grpc-httpjson-transcoding/blob/master/src/json_request_translator.cc, I am not sure how this transcoder work.

jeffsawatzky:
There is a lot of confusion around how `oneOf` in JSON Schema maps to `oneof` in protobufs. There is a long discussion here:
https://github.com/google/gnostic/issues/251

Basically:
`oneOf` in JSON Schema is "this object can be either either SchemaA or SchemaB"
`oneof` in Protobuf is "this Schema can have either propertyA or propertyB"

JSON Schema's `oneOf` more closely maps to using a `google.protobuf.Any` with a restriction on the `@type` param to the types that you allow.

both Envoy and grpc-gateway follow Protobuf's definition of `oneof`, so this change would not work for either of those.

jeffsawatzky:
Also, how would this work if the message and more than one `oneof` in it?

jeffsawatzky:
@ppaanngggg I think what you are actually trying to do is something more like:
```
message Message {
  string message_id = 1;

  message Body {
    string body_id = 1;
  }

  google.protobuf.Any content =2 [(openapi.v3.property) = {
    // use something from #334 to restrict the `@type` to either a `string` or `Message.Body`
  }];
}
```

jeffsawatzky:
Also @ppaanngggg you should just be using the [protojson](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson) marshaller which handles marshalling protobufs correctly based on the [docs](https://developers.google.com/protocol-buffers/docs/proto3#json).

morphar:
I think you have summed it all up @jeffsawatzky 👍

ppaanngggg:
> Basically:
> `oneOf` in JSON Schema is "this object can be either SchemaA or SchemaB"
> `oneof` in Protobuf is "this Schema can have either propertyA or propertyB"

Yes, I just found this problem. I can only support single oneof in message in fact. Thanks for your references, I will learn them.

ppaanngggg:
> Also, how would this work if the message and more than one `oneof` in it?

Yes, you are right. It is impossible to support more than one proto `oneof` in openapi.

I wanted to wrap the name of `oneof` out of fileds, but it is not the correct way to marshal or unmarshal.

Issue #348, opened 2022-04-25
New files not tracked by git (and can we settle on package name?)


jeffsawatzky:
@timburks with the changes in #344, when I run `make` on a clean copy of the repo I get some new files that aren't checked in, and the old files with the old package names still stick around.

See attached screenshot, but in short I have a new `annotations.pb.go` and `OpenAPIv3.pb.go` files under a new `github.com` folder, and the older versions of these files are still checked in to the `openapiv3` folder as well. This might cause some confusion as to which files `protoc-gen-openapi` (and other tools) use. If I delete the older files then `make` no longer works as things are still using the old ones.

![image](https://user-images.githubusercontent.com/421404/165132810-4f0bdc01-6c42-4e73-86f3-d474d9a12b62.png)

I'm not a go expert, and I don't know what the best practices are for package naming and such, but I think some care should be put into thinking about what the package names for these things should be, and ideally solve #337 (the buf.build issue) as well.

I also wonder about the change to `OpenAPIv3.proto` here. From what I understand that file is autogenerated, so does that mean these changes will be reverted the next time the file is autogenerated? Should the thing that autogenerates this file ALSO be updated to fix the package name?

jeffsawatzky:
@timburks I put up #349 which I think fixes the issue with running `make` (or more specifically `COMPILE-PROTOS.sh`).

However it doesn't address the package naming issue between the package name here and the one for the files on buf.build.

jeffsawatzky:
I think this can be closed as the only outstanding issue is the package path, but I think that is better discussed in #337

PR #349, opened 2022-04-25
Fix COMPILE_PROTOS.sh

jeffsawatzky:
This fixes the extra files when running `make` or `COMPILE-PROTOS.sh` as mentioned in #348.


timburks (approved):
@jeffsawatzky Thanks! Does this new go_opt setting cause the .pb.go files to be generated in their existing locations?

jeffsawatzky:
@timburks yes, the .pb.go files now get generated in their previously existing location.

PR #355, opened 2022-05-06
[+] support integer64 as string option

ppaanngggg:
as https://developers.google.com/protocol-buffers/docs/proto3#json

In fact, int64, fixed64, uint64 should be string in json.

So add option to turn int64, fixed64, uint64 to string in openapi


cmd/protoc-gen-openapi/main.go
```diff
@@ -35,6 +35,7 @@ func main() {
 		EnumType:        flags.String("enum_type", "integer", `type for enum serialization. Use "string" for string-based serialization`),
 		CircularDepth:   flags.Int("depth", 2, "depth of recursion for circular messages"),
 		DefaultResponse: flags.Bool("default_response", true, `add default response. If "true", automatically adds a default response to operations which use the google.rpc.Status message. Useful if you use envoy or grpc-gateway to transcode as they use this type for their default error responses.`),
+		Integer64Type:   flags.String("integer64_type", "integer", `type for int64/uint64/sint64/fixed64/sfixed64. Use "string" for string format`),
```
morphar, line 38: Just nitpicking.
Would `int64_type` or `64bit_types` be a better name?

cmd/protoc-gen-openapi/plugin_test.go
```diff
@@ -258,3 +259,40 @@ func TestOpenAPIDefaultResponse(t *testing.T) {
 		})
 	}
 }
+
+func TestOpenAPIStringInteger64(t *testing.T) {
+	for _, tt := range openapiTests {
+		fixture := path.Join(tt.path, "openapi_string_integer64.yaml")
+		if _, err := os.Stat(fixture); errors.Is(err, os.ErrNotExist) {
+			if !GENERATE_FIXTURES {
```
morphar, line 267: Doe this have any relevance to `gnostic`? Or is it something used locally?

cmd/protoc-gen-openapi/plugin_test.go
```diff
@@ -258,3 +259,40 @@ func TestOpenAPIDefaultResponse(t *testing.T) {
...
+			err := exec.Command("protoc",
+				"-I", "../../",
+				"-I", "../../third_party",
+				"-I", "examples",
+				path.Join(tt.path, tt.protofile),
+				"--openapi_out=integer64_type=string:.").Run()
+			if err != nil {
+				t.Fatalf("protoc failed: %+v", err)
+			}
+			if GENERATE_FIXTURES {
```
morphar, line 282: Same question as above.

morphar (commented):
Thanks @ppaanngggg looks good 👍

Maybe 64 bit types should default to `string` as described in the link.
What are your thoughts on this @timburks?

cmd/protoc-gen-openapi/main.go
ppaanngggg: aha, let @timburks decide

cmd/protoc-gen-openapi/plugin_test.go
ppaanngggg: sorry, I don't understand, I just copy from other unit tests

ppaanngggg:
> Maybe 64 bit types should default to `string` as described in the link.

I think it should be in the future. Because `protojson` works like this.

cmd/protoc-gen-openapi/plugin_test.go
morphar: Sorry. I thought `GENERATE_FIXTURES` was new.

cmd/protoc-gen-openapi/main.go
timburks: Do we need a flag for this? I think the [doc you linked](https://developers.google.com/protocol-buffers/docs/proto3#json) should be the source-of-truth for both us and Envoy/grpc-gateway. I'm comfortable with making this fix our only way of handling this.

timburks:
@ppaanngggg Thanks for catching this! I would prefer to just make this the correct behavior and not have an option. @morphar do you need the old behavior?

ppaanngggg:
@timburks I think it should be default.

I just concern about it will break something, afterall I am not familiar with other things in this project.

morphar:
> @morphar do you need the old behavior?

@timburks no. I will always go for "most correct behavior".
It seems that envoy is extremely strict with regards to interpreting int64 as strings as well:
https://github.com/protocolbuffers/protobuf/issues/5015

timburks:
@ppaanngggg let's please make this fix the new behavior (with no option). The link @morphar cited is a good example for us to follow in keeping things simple and using https://developers.google.com/protocol-buffers/docs/proto3#json as our standard.

@ppaanngggg thanks for finding and fixing this!

ppaanngggg:
> Envoy may allow int values as input for int64.

According to doc before, protojson allow both integer and string input, but only string output.

@timburks Great, I will fix it.

cmd/protoc-gen-openapi/generator/reflector.go
```diff
@@ -189,11 +189,13 @@ func (r *OpenAPIv3Reflector) schemaOrReferenceForField(field protoreflect.FieldD
...
-		protoreflect.Int64Kind, protoreflect.Sint64Kind, protoreflect.Uint64Kind,
-		protoreflect.Sfixed32Kind, protoreflect.Fixed32Kind, protoreflect.Sfixed64Kind,
-		protoreflect.Fixed64Kind:
+		protoreflect.Sfixed32Kind, protoreflect.Fixed32Kind:
 		kindSchema = wk.NewIntegerSchema(kind.String())
 
+	case protoreflect.Int64Kind, protoreflect.Sint64Kind, protoreflect.Uint64Kind,
+		protoreflect.Sfixed64Kind, protoreflect.Fixed64Kind:
+		kindSchema = wk.NewStringSchema()
+
```
timburks, line 198: This is great - such a clear and simple change! Thanks!

timburks (approved):

PR #366, opened 2022-08-31
Draft: Handle google/protobuf/wrappers.proto types

galihputera:
TODO


chaayac (approved):

chaayac:
Bump on this. Can a maintainer please review?

galihputera:
up

bickyeric (approved):

galihputera:
I signed it!

morphar:
Hi @galihputera

Sorry for the slow response.
This looks good! Thank you!

Have you tested this against Envoy or something like that?
I don't have a test setup running, but I will get one up and test this with Envoy, just to verify that it converts correctly.

same-id:
> Hi @galihputera
> 
> Sorry for the slow response. This looks good! Thank you!
> 
> Have you tested this against Envoy or something like that? I don't have a test setup running, but I will get one up and test this with Envoy, just to verify that it converts correctly.

Hey,

This has nothing to do with envoy.
Envoy works with the json representation of the protobuf which by design has to work that way.
(i.e. you don't need to set up envoy - just invoke "to_json" on the protobuf and you'll see that these values are translated directly into strings)

You can also see it in the official documentation of wrappers.proto:
https://github.com/google/gnostic/blob/3751138311725a0e9d353b805b2e121c94155d18/third_party/google/protobuf/wrappers.proto#L111
```
// The JSON representation for `StringValue` is JSON string.

```

aditanase:
👍 this looks great - I came here looking for the `Timestamp` conversion - so glad I found this PR.

Haven't tested the wrapper types with envoy - but I can confirm that for `Timestamp` <> `String` it works as advertised, we've been using it in our GRPC APIs for years, with the official ESPv2 json-grpc transcoder.
https://github.com/google/gnostic/blob/3751138311725a0e9d353b805b2e121c94155d18/third_party/google/protobuf/timestamp.proto#L108-L147

Same is valid for `Duration`.

timburks (approved):

galihputera:
Thanks, @timburks for your approval! but I still couldn't merge it. Could you help me do that?

PR #385, opened 2023-04-11
surface: Set status and content-type in response fields

emcfarlane:
Ensures both V2 and V3 derived APIs include the content/type values in the surface fields names like '200 application/json'.

Fixes: https://github.com/google/gnostic/issues/109

This problem is described in more detail in the issue. I've chosen to merge the status code and content type into the name field for both. Current behaviour ignores content type in V2 and adds content type as the name in V3 as a subset of the status code response. This new PR breaks the naming convention to unify both fields, is there a spec for the surface model apis?


timburks:
@emcfarlane Thanks for looking at this. There's no spec for the surface model and limited usage, though I wouldn't be opposed to uses and improvements. It was intended for use in [gnostic-go-generator](https://google/gnostic-go-generator) (now archived) and [gnostic-grpc](https://google/gnostic-grpc), which has limited (but perhaps some) interest.

timburks (approved):

PR #389, opened 2023-04-28
fix: Prevent panic when parsing empty documents

twz123:
The `PaseDocument` functions were returning the first element of the content slice unconditionally, which panicked if the given document bytes were empty. Check if the slice has some elements and return an error if it's empty instead.

See also:
* kubernetes/kubernetes#117363


timburks (approved):
Thanks!

twz123:
The CI failed with

> This request was automatically failed because there were no enabled runners online to process the request for more than 1 days.

Seems to be an intermittent failure of GitHub Actions?

Issue #351, opened 2022-04-27
protoc-gen-openapi: google.protobuf.Duration not string


jan-sykora:
Why is the google.protobuf.Duration converted into OpenAPI as a custom object and not as a string (the JSON format)? 
I would expect it to be represented by `type: string`. Especially when the google.protobuf.Timestamp (which holds the same structure as google.protobuf.Duration) is converted into string (JSON format) as well.

For instance:
```proto
message Baz {
  google.protobuf.Timestamp baz_time = 1;
  google.protobuf.Duration baz_duration = 2;
}

// package google.protobuf
message Timestamp {
  int64 seconds = 1;
  int32 nanos = 2;
}

// package google.protobuf
message Duration {
  int64 seconds = 1;
  int32 nanos = 2;
} 
```
is converted by protoc-gen-openapi into
```yaml
openapi: 3.0.3
components:
  schemas:
    Baz:
      properties:
        bazTime:
          type: string
          format: date-time
        barDuration:
          $ref: '#/components/schemas/Duration'
    Duration:
      type: object
      properties:
        seconds:
          type: integer
          format: int64
        nanos:
          type: integer
          format: int32
```

See [google.protobuf.Duration JSON description](https://github.com/protocolbuffers/protobuf/blob/ece5ef6b9b6fa66ef4638335612284379ee4548f/src/google/protobuf/duration.proto#L94) and [google.protobuf.Timestamp JSON description](https://github.com/protocolbuffers/protobuf/blob/ece5ef6b9b6fa66ef4638335612284379ee4548f/src/google/protobuf/timestamp.proto#L110) for more info.

timburks:
That seems like an oversight. We probably just need to hook up something like [this](https://github.com/google/gnostic/blob/cece63d6719cdc815dae96a2a0b48c7cf693dedb/cmd/protoc-gen-openapi/generator/wellknown/schemas.go#L105).

nightlyone:
FYI: @timburks I took a stab at this in PR #401 and welcome any feedback you have.

PR #401, opened 2023-08-06
fix: Stop lying how GRPC renders durations

nightlyone:
by reporting the data types and formats that are actually used for [google.protobuf.Duration](https://github.com/protocolbuffers/protobuf/blob/ece5ef6b9b6fa66ef4638335612284379ee4548f/src/google/protobuf/duration.proto)

Fixes: #351


timburks (approved):
@nightlyone thanks, this is great.  It adds more detail than we have for the other well-known types, but I think we should just take that as a call to improve the others.

PR #409, opened 2023-10-15
feat: Support format for 64-bit integer field types

esraaghanemali:


arthaali:
@timburks Hello, can you please take a look at this addition to the type `int64`

PR #431, opened 2024-03-07
Revert "[+] support integer64 as string option (#355)"

ajorgensen:
ref: https://github.com/google/gnostic/pull/355
closes: https://github.com/google/gnostic/issues/411

This reverts commit c62333b7964941ab26259ef3648690e6bdd34459.

The original change conflated generating an OpenAPI v3 spec file with the guidance for encoding protobuf messages as json objects. OpenAPI v3 explicitly allows for integer types of int64 (https://swagger.io/docs/specification/data-models/data-types/#numbers). The OpenAPI v3 spec generated by this tool will then usually be fed into language specific generators which will determine how they want to handle each of the types depending on the specific target language.

### DIFF
--- a/cmd/protoc-gen-openapi/examples/tests/mapfields/openapi.yaml
+++ b/cmd/protoc-gen-openapi/examples/tests/mapfields/openapi.yaml
@@ -42,7 +42,8 @@ components:
             type: object
             properties:
                 id:
-                    type: string
+                    type: integer
+                    format: int64
                 label:
                     type: string
         GoogleProtobufAny:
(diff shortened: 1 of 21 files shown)


n0v1:
I also think that the generated OpenAPI specification should match the types in the proto file as close as possible and should not be language-specific. So for proto type `int64` I would expect `{"type": "integer", "format": "int64"}` as it was before instead of `{"type": "string"}.

beihai23 (approved):

suncle1993:
Is there any update in this bug?

ajorgensen:
It appears there is a v2 of the protoc-gen-openapi generator which fixes this issue. Perhaps this project is no longer maintained? I'm going to close this as there appears to be no traction on the issue and https://github.com/grpc-ecosystem/grpc-gateway/tree/main/protoc-gen-openapiv2 appears to work correctly.

PR #459, opened 2025-09-26
Generate int64 types as integer instead of string in OpenAPI schema

### LINKED ISSUE #411: protoc-gen-openapi  BUG
你们的v0.7.0 貌似有问题， 更新之后生成的字段， int64 全变成了string

ajorgensen:
# Description of the Bug

Protobuf int64, sint64, uint64, sfixed64, and fixed64 types were incorrectly being generated as string type in OpenAPI schemas. OpenAPI specification supports int64 as integer type with format specifier, so this change moves 64-bit integer types from string generation to integer generation with appropriate format.

This appears to have been a regression between version v0.6.9 and v0.7.0 of the protoc-gen-openapi generator.

The current code only treats 32 bit integer types as an integer but 64 bit integers as a raw string, this is against the openapi v2 and v3 spec. See the Data Types sections for both versions 

https://swagger.io/specification/v2/ 
https://swagger.io/specification/v3/

Screenshot 2025-09-26 at 11 10 36 AM

This bug results in incorrectly generated openapi files where int64s are written into the specs as being raw strings. This has the the downstream effect of any generated client will treat the integers as string values and require special parsing by the user of the client to parse the long value out of the string.

Examples of the correct values are in the fixture but here is a snippet of what the change in the generate schema looks like for int64:

```diff
 - name: int64_value_type
   in: query
   schema:
-    type: string
+    type: integer
+    format: int64
```

Fixes #411

### DIFF
--- a/cmd/protoc-gen-openapi/examples/tests/additional_bindings/openapi_default_response.yaml
+++ b/cmd/protoc-gen-openapi/examples/tests/additional_bindings/openapi_default_response.yaml
@@ -6,39 +6,35 @@ info:
     title: Messaging API
     version: 0.0.1
 paths:
-    /v1/messages/{messageId}:
-        get:
+    /v1/messages:
+        patch:
             tags:
                 - Messaging
-            operationId: Messaging_GetMessage
-            parameters:
-                - name: messageId
-                  in: path
-                  required: true
-                  schema:
-                    type: string
-                - name: userId
-                  in: query
-                  schema:
-                    type: integer
-                    format: uint64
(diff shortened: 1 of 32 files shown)


ajorgensen:
Evidently this is due to a limitation with Javascript and its handling int64s https://github.com/OAI/OpenAPI-Specification/issues/704. It seems jsonpb specifically will treat int64 and a string for this reason.

rauanmayemir:
Javascript may have a limitation, but there are other platforms that would benefit from int64 hints.

PR #477, opened 2026-07-27
feat: handle response_body option.

jozuenoon:
**Add response_body support to protoc-gen-openapi**

The `response_body` field in `google.api.http` was being ignored. When set, the response schema now uses the named field's type instead of the full output message, matching how body already works for requests.

Scalar/non-message field kinds are logged as unsupported (same as the request side). Missing field names also log a warning.

Added test fixture under `examples/tests/responsebody/` covering a message-typed response body field.


jozuenoon:
This PR covers changes in https://github.com/google/gnostic/pull/406

